PR #20, opened 2025-11-01
Interactive Tutorial For Adding Synthetic Monitoring Plugin

TheComeUpCode:
When should users see your guide? 
If they are interested in Synthetic Monitoring

What are they looking at in the UI when it appears as a recommendation? 
Homepage with navigation bar open

Who should see your guide? 
Everybody

### DIFF
--- a/how-to-setup-synthetic-monitoring/unstyled.html
+++ b/how-to-setup-synthetic-monitoring/unstyled.html
@@ -15,17 +15,23 @@
     <ul>
       <li class="interactive" 
           data-targetaction="highlight" 
-          data-reftarget="a[data-testid='Nav menu item'][href='/admin']"
+          data-reftarget="a[data-testid='data-testid Nav menu item'][href='/admin']"
           data-requirements="navmenu-open,exists-reftarget">
         Open the main menu and click <strong>Administration</strong>
       </li>
       
       <li class="interactive" 
           data-targetaction="highlight" 
-          data-reftarget="a[data-testid='Nav menu item'][href='/admin/plugins']"
-          data-requirements="navmenu-open,exists-reftarget">
-        Click <strong>Plugins</strong>
+          data-reftarget="a[data-testid='data-testid Nav menu item'][href='/admin/plugins']"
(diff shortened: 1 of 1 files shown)


how-to-setup-synthetic-monitoring/unstyled.html
```diff
@@ -0,0 +1,71 @@
...
+          data-targetaction="highlight" 
+          data-reftarget="a[data-testid='data-testid Nav menu item'][href='/admin']"
+          data-requirements="navmenu-open,exists-reftarget">
+        Open the main menu and click <strong>Administration</strong>
+      </li>
+      
+      <li class="interactive" 
+          data-targetaction="highlight" 
+          data-reftarget="a[data-testid='data-testid Nav menu item'][href='/admin/plugins']"
+          data-requirements="exists-reftarget">
```
moxious: requires navmenu-open, since you can't click on the navmenu item without it.

how-to-setup-synthetic-monitoring/unstyled.html
```diff
@@ -0,0 +1,71 @@
...
+
+  <span id="install-synthetic-monitoring" class="interactive" data-targetaction="sequence" data-reftarget="span#install-synthetic-monitoring">
+    <ul>
+      <!-- TODO: the plugins search input does not have an id, so we need to use a different selector, in this case the placeholder text -->
+      <li class="interactive" 
+          data-targetaction="formfill" 
+          data-reftarget="input[placeholder='Search Grafana plugins']"
+          data-targetvalue="Synthetic Monitoring"
+          data-requirements="exists-reftarget">
+        Type "Synthetic Monitoring" in the search bar
```
moxious: Here's a trick.... you have the option to define an extra "comment" that shows up when you highlight a box.

When you highlight the search box and type something into it, and the user clicks "Show Me", it will draw an orange highlight box around the input.   By using an extra span, you can cause text to pop up explaining the element that's being highlighted.

Here's an example:  https://github.com/grafana/interactive-tutorials/blob/main/rca-demo-v2/unstyled.html#L84

PR #23, opened 2025-11-06
SM interactive guides

TheComeUpCode:
Adding the steps to creating the first SM HTTP check.


moxious (approved):
Reviewed against https://dev.grafana-dev.net/

Ran it in several different scenarios, it LGTM.  Added one tiny commit to align a few words with a button label. Thanks for the excellent work

PR #50, opened 2025-12-12
validate guide JSON using zod/pathfinder

moxious:
This PR introduces the use of Zod JSON schema validation from the Pathfinder repo, here for interactive guides.

This is the first step in the wider plan to "treat content like software".  We're doing much more than checking that the JSON parses correctly; checking schema conformance, and also the proper micro-grammar use of things like requirements and objectives.  These same checks are put into Pathfinder, so guides passing strict checks here is necessary for them to load cleanly into pathfinder and be assessable as "safe".

### DIFF
--- a/.cursor/best-practices.mdc
+++ b/.cursor/best-practices.mdc
@@ -50,6 +50,57 @@ Make admin-only steps `skippable: true` with helpful hints.
 Use `objectives` to auto-complete already-done steps.
 Provide alternative workflows for different user permissions.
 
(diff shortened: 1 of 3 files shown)


irm-configuration/content.json
```diff
@@ -78,9 +76,7 @@
     },
     {
       "type": "guided",
-      "id": "create-escalation-chain",
```
moxious: not every block has/allows an ID

tomglenn (approved):

PR #290, opened 2026-04-17
Docs: adds guides required for alerting course

chri2547:
Adds six standalone interactive guides for the Grafana Alerting alert rule creation form, each covering a specific section of the UI: naming, query and condition, folders and labels, evaluation behavior, notifications, and notification messages. These guides use verified selectors from the live UI and are designed to support an upcoming alerting course in the Learning Hub.


chri2547:
Closing because this was an experiment i was working on with Bev. no longer needed.

PR #299, opened 2026-04-24
Add deploy preview for interactive guides

jdbaldry:
This PR adds the website front matter metadata as a YAML file in each guide's package directory.
The website can use this metadata to construct the ordered learning path from a set of interactive guides.

Moving the metadata to this repository allows authors to preview how their content looks in the website without having to worry about a separate repository or how the data syncs between here and there. I believe some of this additional metadata could be derived from the core package metadata but I might save that for a future consolidation. They now can think purely in terms of the interactive-tutorials repo and essentially ignore the website repo aspect of publishing

The only meaningful changes to the workflow are the `SOURCES` configuration which, configures where the content appears in the build, and the removal of the paths filter since we now use labels to decide whether or not to run the workflow.

Requires https://github.com/grafana/writers-toolkit/pull/1383 to be merged and the Writers' Toolkit reference in this workflow to be updated from `jdb/deploy-preview` to `main` before merge.

Closes https://github.com/grafana/website/issues/30386.


moxious (approved):

chri2547 (approved):
Thanks so much!

robbymilo (approved):
Sweet!

tomglenn (approved):

PR #319, opened 2026-05-15
Pathfinder LJs: procedural-only milestones in manifests

### LINKED ISSUE #315: Remove unwanted milestones from LPs
PF manifest allowed us to specify which learning path milestones we want to display in PF. Update the manifest of all published LPs to remove uninteresting (i.e. overview, business value) milestones. In the Pathfinder plugin, user need to jump straight into the steps.

tacole02:
This PR tightens in-product (Pathfinder) learning paths so people hit hands-on steps sooner.

We dropped long “why / value / advantages / case for” style milestones from the Pathfinder journey lists, and trimmed a few planning / “how it works” steps where they were still acting like a slow front door. Integration paths keep the short “understand dashboards and alerts” milestones, because those explain the bundled assets people use right after setup—they’re not the same as the long framing intros we removed.

Why: In the product, we want learners doing (clicking through the guide, using Grafana) as fast as possible. The longer context and narrative still belongs on the web docs version of the paths, where people expect to read and skim. Pathfinder should feel like “start here, do this”, not like reading a chapter before you’re allowed to touch the UI.

---

> [!NOTE]
> **Medium Risk**
> Primarily updates learning-path `manifest.json` milestone lists and `depends`/`recommends` links; risk is limited to potentially breaking or reordering Pathfinder journey flow if any removed milestone IDs are still referenced elsewhere.
> 
> **Overview**
> **Tightens multiple Pathfinder learning paths to be more hands-on.** Across many `*-lj` path/guide `manifest.json` files, this PR removes introductory/value/planning/how-it-works milestones from `milestones` lists and clears corresponding `depends` links so users start directly on procedural steps.
> 
> It also retargets a few step-to-step transitions (for example, making `kafka-monitoring-install-grafana-alloy` depend on `kafka-monitoring-configure-jmx-exporter`, and `private-data-source-connect-deploy-pdc-agent` depend on `...-install-pdc-binary`) and normalizes manifest formatting (expanded JSON objects/arrays for `author`, `testEnvironment`, `recommends`, and `suggests`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc969f864ea02866233f8c23fb60aadf831d15f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/adaptive-logs-lj/review-apply/manifest.json
+++ b/adaptive-logs-lj/review-apply/manifest.json
@@ -3,19 +3,10 @@
   "type": "guide",
   "description": "Learn how to review and apply Adaptive Logs recommendations to reduce log volume while keeping dashboards and alerts working.",
   "category": "data-availability",
-  "author": {
-    "team": "Grafana Documentation",
-    "name": "chri2547"
-  },
-  "testEnvironment": {
-    "tier": "cloud"
-  },
-  "depends": [
-    "adaptive-logs-how-it-works"
-  ],
-  "recommends": [
-    "adaptive-logs-exemptions"
-  ],
+  "author": { "team": "Grafana Documentation", "name": "chri2547" },
+  "testEnvironment": { "tier": "cloud" },
+  "depends": ["adaptive-logs-how-it-works"],
+  "recommends": ["adaptive-logs-exemptions"],
   "suggests": [],
   "provides": []
 }
--- a/adaptive-logs-lj/review-apply/manifest.json
+++ b/adaptive-logs-lj/review-apply/manifest.json
@@ -3,10 +3,19 @@
   "type": "guide",
   "description": "Learn how to review and apply Adaptive Logs recommendations to reduce log volume while keeping dashboards and alerts working.",
   "category": "data-availability",
-  "author": { "team": "Grafana Documentation", "name": "chri2547" },
-  "testEnvironment": { "tier": "cloud" },
-  "depends": ["adaptive-logs-how-it-works"],
-  "recommends": ["adaptive-logs-exemptions"],
+  "author": {
+    "team": "Grafana Documentation",
+    "name": "chri2547"
+  },
+  "testEnvironment": {
+    "tier": "cloud"
+  },
+  "depends": [
+    "adaptive-logs-how-it-works"
+  ],
+  "recommends": [
+    "adaptive-logs-exemptions"
+  ],
   "suggests": [],
   "provides": []
 }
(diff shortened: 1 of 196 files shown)


adaptive-logs-lj/review-apply/manifest.json
```diff
@@ -5,7 +5,7 @@
   "category": "data-availability",
   "author": { "team": "Grafana Documentation", "name": "chri2547" },
   "testEnvironment": { "tier": "cloud" },
-  "depends": ["adaptive-logs-how-it-works"],
+  "depends": [],
```
moxious, line 8: I understand the intent of this PR is to remove those items from the critical path in-product; but I'm not sure it makes sense to sever these depends links.  If you're cutting these links, this basically produces guides leftover that have no context.  If they're truly never going to be used again, then we might want to remove them entirely rather than just removing them from the "milestones" path of the learning journey.  If you think that they ARE going to be used (at least sometimes, by some people) then I think that destroying the graph connections that connects them isn't a good idea; it would be hard work to reconstitute that later if you change your mind.

This is a systematic comment across the PR as this pattern holds in most of your changes.

moxious (changes requested):
This PR is quite a bit larger than it needs to be because most of the work it's doing is just reformatting JSON.   It seems like the real core bits are to remove the first step or two in the "milestones" of learning journey.  That notional change is fine (if approved by writers, that's a content call, I would merge it).  Please take careful note of the "severing the metadata links" comment on here.  That seems to be going further than is necessary for this PR and I would advise against it unless there's extra rationale (eventual removal of those extra steps) you want to post to the PR.

PR could be substantially slimmed by just not reformatting JSON and leaving lines that you aren't modifying alone.

tacole02:
Hi @moxious 

Thanks for the review. 

I put the depends and recommends links back to how they were on main for the intro/plan steps we took off the path, so those guides are no longer disconnected in metadata. Pathfinder still uses the shorter milestones lists, and only the new first step has depends cleared so users are not blocked on a step that is no longer on the journey.

Let me know if this adequately addresses your concerns. Thank you!

adaptive-logs-lj/review-apply/manifest.json
tacole02: I think that's a good point, and these milestones may very well be used again, just not as part of this learning path. 

I've replaced the depends links, except for each path’s new first step (set to []), so Pathfinder does not require a framing or plan milestone that is no longer on the path.

Let me know what you think.

tiffany76 (approved):
LGTM for the categorize-collector-fleet-lj.

moxious (approved):

jtvdez (approved):

jtvdez:
billing and usage changes LGTM

knylander-grafana (approved):
Trusting the other reviewers...

BeverlyJaneJ (approved):

PR #317, opened 2026-05-15
Add interactive guide for configure IRM learning path

alyssawada:
Adds the grafana-irm-configuration-lj learning path Pathfinder package.

Supports an upcoming Pathfinder experiment comparing the standalone IRM interactive guide with the interactive Configure IRM learning path.

Related [Website PR](http://github.com/grafana/website/pull/30670) updating the learning path and wiring the learning path pages to the Pathfinder package

---

> [!NOTE]
> **Low Risk**
> Low risk content-only change: adds new Pathfinder learning path JSON packages and a CODEOWNERS entry, with no runtime logic or security-sensitive behavior changes.
> 
> **Overview**
> Adds a new `grafana-irm-configuration-lj` Pathfinder learning path package with a `path` manifest plus eight milestone guides covering IRM value, component mental model, schedule and escalation setup, Grafana Alerting integration, alert-rule wiring, end-to-end testing, and completion.
> 
> Updates `.github/CODEOWNERS` to assign ownership of the new `grafana-irm-configuration-lj/` directory.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3da6b68a077d41b58ade1d220901f6b3513f895. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #325, opened 2026-05-19
fix(grafana-irm-configuration-lj): render images, callouts, and videos in Pathfinder

Jayclifford345:
## Summary

The `grafana-irm-configuration-lj` learning journey (merged in #317) ships three Pathfinder-incompatible patterns inherited from Hugo-rendered docs. This PR fixes all three so the LJ renders correctly inside Pathfinder.

### What was broken

- **Images don't render.** Four images used website-relative paths (`/media/docs/grafana-cloud/alerting-and-irm/...`) that only resolve on grafana.com. In Pathfinder, the browser resolves them against the Pathfinder host and 404s. Documented in [`.cursor/skills/migrate-guide/SKILL.md`](.cursor/skills/migrate-guide/SKILL.md) (lines around 455).
- **Admonitions display as literal text.** Three blocks wrapped content in Hugo shortcodes (`{{< admonition type=\"...\" >}}...{{< /admonition >}}`). The JSON guide schema has no admonition block type ([`docs/json-guide-reference.md`](docs/json-guide-reference.md)), so the tags pass through verbatim. No other LJ in the repo uses admonitions.
- **Video shortcodes display as literal text.** Two blocks used `{{< docs/video id=\"...\" >}}...{{< /docs/video >}}` — same bug class.

### What changed

| File | Change |
|------|--------|
| `content.json` | Image URL → absolute |
| `build-mental-model/content.json` | Image URL → absolute; video shortcode → `video` block + adjacent markdown |
| `create-on-call-schedule/content.json` | Image URL → absolute; admonition (tip) → `> **Tip:**` blockquote |
| `create-escalation-chain/content.json` | Image URL → absolute; admonition (note) → `> **Note:**` blockquote |
| `run-end-to-end-test/content.json` | Admonition (note) → `> **Note:**` blockquote |
| `understand-value/content.json` | Video shortcode → `video` block + adjacent markdown |

Blockquote callout style was chosen to match the existing convention in `infinity-csv-lj`, `infrastructure-alerting-lj`, and `create-availability-slo-lj`. Video block format matches `docs/json-guide-reference.md:135`.

### Known pre-existing typo (not fixed here)

The asset `screenshot-grafana-irm-escalatino-chain-with-schedule-lj.png` has `escalatino` (should be `escalation`) in the source filename on grafana.com. The reference is correct against the published asset (returns 200), so this PR leaves it alone. Recommend re-uploading the asset with the corrected name in a follow-up change, then updating the reference.

## Test plan

- [x] `jq empty` on every modified JSON file — all parse cleanly
- [x] `grep` for residual website-relative `/media/docs/grafana-cloud/...` refs — none remain
- [x] `grep` for residual `{{<` Hugo shortcodes — none remain
- [x] All four image URLs return `200 OK` at the new absolute path (verified with `curl` during planning)
- [ ] Load LJ in a Pathfinder dev instance and walk every milestone — confirm images render inline, callouts render as blockquotes, both videos embed and play
- [ ] Run Pathfinder CLI validation: `node {pathfinder-app}/dist/cli/cli/index.js validate --packages .`

### DIFF
--- a/grafana-irm-configuration-lj/build-mental-model/content.json
+++ b/grafana-irm-configuration-lj/build-mental-model/content.json
@@ -9,15 +9,21 @@
     },
     {
       "type": "markdown",
-      "content": "The following diagram illustrates the different components of IRM and how they work together as alerts flow through the system:\n\n![Alert flow diagram showing alert source passing through Grafana IRM components](/media/docs/grafana-cloud/alerting-and-irm/screenshot-grafana-irm-alert-flow-diagram-lj.png)"
+      "content": "The following diagram illustrates the different components of IRM and how they work together as alerts flow through the system:\n\n![Alert flow diagram showing alert source passing through Grafana IRM components](https://grafana.com/media/docs/grafana-cloud/alerting-and-irm/screenshot-grafana-irm-alert-flow-diagram-lj.png)"
     },
     {
       "type": "markdown",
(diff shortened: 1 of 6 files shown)


alyssawada (approved):
Thank you for catching these!

Issue #314, opened 2026-05-13
Migrate LPs to avoid no-op workaround


tacole02:
Historically, we've used `noops` in interactive learning paths as a workaround to maintain numbering across interactive/non-interactive steps.

https://github.com/grafana/grafana-pathfinder-app/pull/871 (https://github.com/grafana/grafana-pathfinder-app/issues/841) resolves this issue so that this workaround is no longer necessary. In fact, it's detrimental.

After the above PR is deployed and tested, we need to migrate all existing learning paths to avoid this workaround.

Tracking: https://docs.google.com/spreadsheets/d/1ANrY6fBu42d5g7XiX8zfkRFDcVu48ftu2xC2PtKCzDQ/edit?usp=sharing

PR #333, opened 2026-05-20
Migrate adaptive metrics numbered steps

tacole02:
Migrates the Adaptive Metrics learning path from noop workaround steps to markdown content blocks so Pathfinder can render them as sequentially numbered section steps where applicable. Part of https://github.com/grafana/interactive-tutorials/issues/314.

### DIFF
--- a/adaptive-metrics-lj/auto-apply/content.json
+++ b/adaptive-metrics-lj/auto-apply/content.json
@@ -23,8 +23,7 @@
             ]
         },
         {
-            "type": "interactive",
-            "action": "noop",
+            "type": "markdown",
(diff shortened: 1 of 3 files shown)


tacole02:
Hi @Eve832, quick note while this migration is open: a few of these former `noop` blocks are not inside a `section` block, so converting them to `markdown` removes the no-op workaround but does not make them render as numbered section steps.

That appears to match the previous behavior for those parts of the LJ, since they were already outside a section. I left the structure unchanged in this PR to keep the migration small and focused.

Do you want to fix that separately by wrapping those procedural steps in sections so they render with numbered steps, or should we leave the current structure as-is?

Eve832 (approved):
lgtm!

tacole02:
@brendamuir mind taking a look as codeowner?

brendamuir (approved):

PR #340, opened 2026-05-20
Migrate billing usage numbered steps

tacole02:
Migrates the billing usage learning path from noop workaround steps to markdown content blocks so Pathfinder can render them as sequentially numbered section steps. Part of https://github.com/grafana/interactive-tutorials/issues/314.

### DIFF
--- a/billing-usage-lj/create-billing-alert/content.json
+++ b/billing-usage-lj/create-billing-alert/content.json
@@ -97,14 +97,8 @@
           "doIt": false
         },
         {
-          "type": "interactive",
-          "action": "highlight",
-          "reftarget": "a[data-testid='data-testid Nav menu item'][href='/alerting/routes']",
-          "content": "Notification policies are configured separately from alert rules. If you need to route alerts to different teams based on labels, navigate to **Notification policies** in the sidebar. Refer to the [notification policies documentation](https://grafana.com/docs/grafana-cloud/alerting-and-irm/alerting/configure-notifications/create-notification-policy/) for details.",
-          "requirements": [
-            "navmenu-open"
-          ],
-          "doIt": false
+          "type": "markdown",
+          "content": "Notification policies are configured separately from alert rules. If you need to route alerts to different teams based on labels, use **Notification configuration** and open the **Notification policies** tab. Refer to the [notification policies documentation](https://grafana.com/docs/grafana-cloud/alerting-and-irm/alerting/configure-notifications/create-notification-policy/) for details."
         },
         {
           "type": "interactive",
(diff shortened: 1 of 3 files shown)


jtvdez:
Going through this and the "show me" of step 14 of the create a billing alert path is broken:
```
Notification policies are configured separately from alert rules. If you need to route alerts to different teams based on labels, navigate to Notification policies in the sidebar. Refer to the [notification policies documentation](https://grafana.com/docs/grafana-cloud/alerting-and-irm/alerting/configure-notifications/create-notification-policy/) for details.
```
I think the left-side nav originally had "Notification policies" as an item but they are under "Notification configuration" as a tab now. We could adjust the wording of that step and remove the show me as a quick fix

tacole02:
Good catch. This looks unrelated to the noop-to-markdown migration because that step wasn't changed in this PR, but it is a stale navigation/show-me step. I'll handle it here so this PR also fixes the broken **Notification policies** guidance.

tacole02:
Pushed a fix in 45ec9ba. I removed the stale show-me highlight for the old **Notification policies** sidebar route and changed it to markdown guidance that points users to **Notification configuration** > **Notification policies**. @jtvdez please have another look!

tacole02:
While checking this, I noticed a separate pre-existing issue in this milestone: a few informational blocks are mixed in with the actual user steps, so the step flow is harder to follow than it should be.

For example, in `billing-usage-lj/create-billing-alert/content.json`, the “The following image shows how to navigate to the **New alert rule** menu option from a panel…” block appears in the middle of the step sequence, but it’s explanatory context rather than a user action.

This wasn’t introduced by this PR. The migration only updated the existing workaround blocks, and those blocks are actual steps.

@jtvdez Could you open a follow-up PR to clean up the structure here? Thank you!

Eve832:
I'm working on redoing this LP in accordance with the new billing and usage app, fwiw

tacole02:
@Eve832 could we still go ahead and push this change? It's just migrating non-interactive steps from noop to markdown block type now that consistent step numbering is supported.

jtvdez:
@tacole02 I'll defer to @Eve832 but your changes here look fine from my end -  if it's worth me opening a follow up, I'm happy to do it but might not be the best move if there are already plans to redo this LP.

tacole02:
Thanks, @jtvdez ! I agree it doesn't make sense to open a follow-up if Eve is planning on redoing it anyways. @Eve832 approval is in your hands 😄

Eve832 (approved):
👍

PR #343, opened 2026-05-21
Formalize shared/snippets/ as the live-reference snippet system

tomglenn:
## Summary

Establishes the shape the Pathfinder app will consume via `snippet-ref` blocks. Companion PR: grafana/grafana-pathfinder-app#907.

The legacy `shared/index.json` was an unfinished proposal that combined templates and snippets in an array form and was never actually deployed. This PR replaces it with:

- **`shared/snippets/index.json`** — map keyed by ID, generated from bodies. This is what the Pathfinder app fetches at `<host>/shared/snippets/index.json`.
- **`description` is now required** on every snippet body. Without it the picker UI in the editor is unbrowsable.
- **`scripts/build-snippets-index.py`** — regenerates the index from bodies. The body is the source of truth; the index is derived. CI runs the same script in check mode and fails on drift.
- **`shared/snippets/README.md`** — documents the layout for snippet authors.

Templates (`shared/templates/tutorial-datasources.json`) are **untouched**. Templates are a distinct concept (copy-as-starting-point, not live reference) and will get their own design pass.

## Migration notes

- 12 snippets that were in the old `shared/index.json` keep their existing descriptions, now stored on the body.
- 8 snippets that existed as files but weren't in the old index get new one-sentence descriptions written for them: `expand-collapse-arrow`, `metrics-drilldown-nav`, `metrics-drilldown-tabs`, `tab-button-pattern`, `traces-drilldown-nav`, `traces-drilldown-tabs` (plus `panel-menu-explore` and `save-dashboard-button` which already had descriptions on their bodies).
- `shared/snippets/case-for-o11y/` (directory-style snippet, never indexed) is **left as-is** — out of scope for this PR.

## Coordination

This is paired with grafana/grafana-pathfinder-app#907 which adds the consumer side. Both should land together — the consumer renders the error placeholder until this lands, this is dormant until the consumer ships.

## Test plan

- [ ] CI: the new `validate-snippets-index` job passes.
- [ ] CI: existing JSON validation jobs still pass.
- [ ] Once Pathfinder app PR is in: load the catalog from a dev/preview CDN, confirm 20 entries appear in the picker.
- [x] Local: `python3 scripts/build-snippets-index.py` regenerates the index without diff.
- [x] Local: every snippet body has `id`, `title`, `description`, `blocks`.


moxious (changes requested):
Nice cleanup — the source-of-truth pattern and the validate-on-PR job are exactly right. One blocking change before I can approve, plus a couple of optional follow-ups.

**Blocking**

`scripts/build-snippets-index.py` needs to enforce `blocks` as a required field. Right now `REQUIRED = ("id", "title", "description")` — a body shipped without `blocks`, or with `blocks: []`, would pass CI and land in the catalog as a broken entry. The README already documents `blocks` as required, so this is just bringing the validator in line with the contract.

Suggested change:

```python
REQUIRED = ("id", "title", "description", "blocks")

# ...inside the loop, after the REQUIRED check:
if not isinstance(body.get("blocks"), list) or not body["blocks"]:
    errors.append(f"{filename}: 'blocks' must be a non-empty array")
    continue
```

**Optional (happy to merge without, but worth considering)**

1. **README claim vs. enforcement.** The README states `blocks` "cannot contain `snippet-ref` — snippets cannot reference snippets." If that's a hard rule, the builder should reject it; if it's aspirational, soften the wording. Either is fine — just pick one so authors aren't surprised later.

2. **Duplicate-id defense.** Since `id` must match the filename and filenames are unique on disk, duplicate ids are structurally impossible today. A one-line `if body["id"] in index:` check would make the failure mode explicit if that invariant ever loosens.

3. **Tiny readability nit:** `filename[:-len(".json")]` → `os.path.splitext(filename)[0]`.

Everything else looks good — deploy ordering is correct, the `.gitignore` entry is scoped tightly, the migration notes are clear about which snippets got new descriptions, and `case-for-o11y/` being skipped by the directory filter is the right call.

PR #368, opened 2026-05-26
ci(deploy): isolate pathfinder-app CLI from GCS credentials

Jayclifford345:
## Summary

Splits `deploy.yml` into three jobs so third-party code checked out from `grafana/grafana-pathfinder-app@main` never runs in a job that holds the GCS push credential.

- **`deploy-guides`** — trusted; holds `id-token: write`; only runs first-party shell commands and SHA-pinned Grafana shared-workflow actions to publish `guides/`.
- **`build-packages`** — sandboxed; **no `id-token: write`**; runs the pathfinder CLI to produce `packages/repository.json`, `packages/graph.json`, and `courses/*.json`, then uploads them as artifacts.
- **`publish-packages`** — trusted; holds `id-token: write`; `needs: build-packages`; downloads the artifacts and pushes `packages/` and `courses/` to GCS via the SHA-pinned `push-to-gcs` action.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -22,15 +22,17 @@ on:
 
 permissions:
   contents: read
-  id-token: write
 
 env:
   DEPLOY_ENV: ${{ github.event_name == 'push' && 'prod' || github.event.inputs.environment }}
 
 jobs:
-  deploy:
-    name: Deploy to GCS
+  deploy-guides:
+    name: Deploy guides/ to GCS
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
     steps:
       - name: Checkout
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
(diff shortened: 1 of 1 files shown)


moxious (changes requested):
## Requested changes

Before this can land, please address the following:

### 1. Rebase — checkout action is downgraded
This PR pins `actions/checkout@34e1148… # v4.3.1` on all three jobs, but `main` is now on `df4cb1c… # v6.0.3` (Renovate merged in `7b4462d`). Rebase and update the SHAs/version comments so the merge doesn't silently revert the bump.

### 2. Courses failure path is inconsistent and will break `publish-packages`
The chain is currently:
- `build-courses` — `continue-on-error: true` (failure tolerated)
- `Upload courses artifact` — `if-no-files-found: warn` (no artifact uploaded on empty output)
- `publish-packages` → `Download courses artifact` — **fails hard** if the artifact doesn't exist

So a soft failure in `build-courses` now hard-fails the publish job, which is a regression from current behavior. Pick one and apply it end-to-end:

- **(a)** Treat courses as optional: add `continue-on-error: true` to the download step and gate the courses push with `if: ${{ hashFiles('courses/**') != '' }}`, or
- **(b)** Treat courses as required: set `if-no-files-found: error` on upload and drop `continue-on-error` from `build-courses`.

### 3. Move validation into the trusted job
`Validate packages before publish` currently runs inside `build-packages`, i.e. it is executed by the same untrusted `pathfinder-app@main` CLI that produced the artifacts — so it cannot be trusted to gate publish. Please re-run validation (or a minimal integrity check) inside `publish-packages` after `download-artifact` but before `push-to-gcs`. Without this, the sandboxing only mitigates credential exfiltration, not artifact tampering.

### 4. Restore the `EXCLUDE` rationale comment
The block comment explaining why each entry in `EXCLUDE="pathfinder-app .github docs .cursor shared guides packages courses"` is excluded was removed. That list is not self-evident; please put the rationale back.

---

Once 1–5 are addressed I'm happy to re-review. The overall split is the right move — these are just the deltas needed to land it safely.

moxious (approved):

leslie-heinzen (approved):
LGTM :shipit:

PR #374, opened 2026-05-31
[codex] Add DPM explore up metric guide

jmichalek132:
## What changed

- Added a standalone `dpm-explore-up-metric` guide with `content.json` and `manifest.json`.
- Guides users through Explore with the instant query `sort_desc(count_over_time({__name__="up"}[1m]))`.
- Adds advanced exact-name and regex query examples with a query cost caveat.
- Links to the Grafana Cloud DPM docs, the DPM blog post, and `grafana-ps/dpm-finder`.

## Why

This gives Pathfinder a simple first guide for estimating data points per minute before moving on to more complex flows.

## Validation

- `node /Users/jorge/git/work/grafana-pathfinder-app/dist/cli/cli/index.js validate --package dpm-explore-up-metric`
- `node /Users/jorge/git/work/grafana-pathfinder-app/dist/cli/cli/index.js build-repository . --exclude shared -o /private/tmp/interactive-tutorials-repository.json`

Tested in my stack
Screenshot 2026-06-01 at 15 42 29
Screenshot 2026-06-01 at 15 42 24

## Notes

- No local PR template was found in `grafana/interactive-tutorials`.
- Manual preview should verify the live Explore selectors, especially the manual **Type** to **Instant** step.

PR #389, opened 2026-06-13
fix(dpm-explore-up-metric): correct recommender targeting

jmichalek132:
## What

Fixes the `targeting` block in `dpm-explore-up-metric/manifest.json` so the guide is actually surfaced where it's relevant.

## Why

Follow-up to #374. The original targeting recommended the guide on `/billing` and `/usage`, which are **not real Grafana routes**. The recommender's `urlPrefix` match is a `startsWith` test against `location.pathname`, so those prefixes never matched any page and the guide could never surface there. `/dashboards` was also too broad for a DPM-in-Explore flow.

(Flagged by @ in the [review thread](https://github.com/grafana/interactive-tutorials/pull/374): "you're recommending yourself on /billing which doesn't exist".)

## Change

| Before | After |
| --- | --- |
| `/explore` | `/explore` (cross-platform) |
| `/dashboards` | _(dropped)_ |
| `/billing` | `/a/grafana-costmanagementui-app` (cloud-only) |
| `/usage` | _(folded into the above)_ |

```json
"targeting": {
  "match": {
    "or": [
      { "urlPrefix": "/explore" },
      { "and": [
          { "targetPlatform": "cloud" },
          { "urlPrefix": "/a/grafana-costmanagementui-app" }
      ]}
    ]
  }
}
```

- **Explore** (`/explore`) stays cross-platform: it's where the guide tells you to run the DPM query.
- **Cost management / billing** is the Grafana Cloud cost-management app at `/a/grafana-costmanagementui-app`. The no-trailing-segment prefix covers the overview and every sub-tab (`/metrics`, `/logs`, `/traces`, `/profiles`). Gated with `targetPlatform: cloud` since the app is Cloud-only — this mirrors the existing `billing-usage-lj` journey's targeting.

### DIFF
--- a/dpm-explore-up-metric/manifest.json
+++ b/dpm-explore-up-metric/manifest.json
@@ -14,13 +14,14 @@
           "urlPrefix": "/explore"
         },
         {
-          "urlPrefix": "/dashboards"
-        },
-        {
-          "urlPrefix": "/billing"
-        },
-        {
-          "urlPrefix": "/usage"
+          "and": [
+            {
+              "targetPlatform": "cloud"
+            },
+            {
+              "urlPrefix": "/a/grafana-costmanagementui-app"
+            }
+          ]
         }
       ]
     }


moxious (approved):

moxious:
@jmichalek132 due to new policy, all commits must be signed. Branch needs to be revised, then I can re-approve and merge

PR #392, opened 2026-06-17
Adds new Azure monitoring interactive guide

HCloward:
I used the cursor command to create this new guide and tested the steps to be sure they work. It seems fine to me, but I haven't had an interactive guide merged before so not sure what next steps are.


Eve832:
Some suggestions:
- for step one, you can use the `navigate` action and just go directly to Azure if you want, or use it to break up the navigation if discovery is super important. Leaving it as-is takes a long time to run
- running step 1 does not unlock step two because it doesn't take you all the way to the Azure page
That's as far as could go

HCloward:
Closing this PR in favor of a new one here: https://github.com/grafana/interactive-tutorials/pull/403

PR #429, opened 2026-07-12
Guide: Get started with Grafana Cloud

chri2547:
New Grafana Cloud users land in a broad product with no single obvious first step, and many drop off before reaching a first success. This PR adds a Pathfinder "Get started" hub that asks users what they want to do first and routes them to the matching hands-on path, so they can reach a quick win.

- `get-started-hub` — a router guide. Each door is a `navigate` step that deep-links to the relevant page and opens the matching guide via `openGuide`.
- `get-started-explore` — see Grafana Cloud in action with demo dashboards (no data setup).
- `get-started-connect-data` — query data where it lives by adding a data source.
- `get-started-monitor-host` — monitor a host or service by installing an integration.

The hub also routes to three existing journeys: 
* Run a load test with k6
* Monitor uptime with Synthetic Monitoring
* Send metrics from an existing collector (Prometheus remote write) 

All six first-success paths live behind one entry point. Each spoke ends with a link back to the hub so users can pick another path.


chri2547:
We shall revisit the hub-and-spoke concept after PLG work solidifies.

PR #443, opened 2026-07-15
Adds a tour of the Eurovision dashboard on Play.

simonprickett:
This PR adds a tour of the [Eurovision dashboard](https://play.grafana.org/d/eurovision/eurovision-2026) that's new to Play.

### DIFF
--- a/play-eurovision-tour/content.json
+++ b/play-eurovision-tour/content.json
@@ -1,0 +1,10 @@
+{
+  "id": "play-eurovision-tour",
+  "title": "Take a Tour of the Eurovision 2026 Dashboard",
+  "blocks": [
+    {
+      "type": "markdown",
+      "content": "## Introduction\n\nThis dashboard provides an overview of Eurovision 2026.\n\n_TODO: Replace this introduction once the dashboard JSON is provided. Describe what the dashboard shows and the Grafana features it demonstrates._"
+    }
+  ]
+}
--- a/play-eurovision-tour/content.json
+++ b/play-eurovision-tour/content.json
@@ -4,7 +4,266 @@
   "blocks": [
     {
       "type": "markdown",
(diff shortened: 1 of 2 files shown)


leslie-heinzen:
Content itself looks good to me 👍🏻 But the guide fails E2E testing.

I ran this against play.grafana.org with the Pathfinder E2E runner: **10 of 26 steps pass**, aborting at step 11. Three changes needed.

**1. Drop `completeEarly: true` from both guided blocks.** On a `guided` block this marks the block complete *before any substep runs*, so the runner skips the substeps and reports a pass for work never done. The panel editor never opens, which is the real reason step 11 fails. This is a known pathfinder plugin defect; for now, avoid the completeEarly flag on guided blocks (it's still valid on single-action `interactive` blocks).

**2. Move the Table view toggle out of the `guide-section-sql` guided block** into its own `interactive` step. Once the completeEarly flag is gone, that block drives all 4 substeps and then hangs indefinitely. Guided completion doesn't resolve for the table-view switch.

**3. Retarget 5 selectors.** The **new query editor** is used, which doesn't expose `data-testid Query editor row` at all, so these can never resolve:
- `Query editor row:nth-match(1)` → `button[aria-label="Show query"]` (×2, the hidden Google Sheets query)
- `Query editor row:nth-match(2)` → `button[aria-label="Hide expression"]` (×3, the SQL expression)

No content text needs to change — the new labels match what each step already describes.

With those three, the guide passes 27/27 (I verified locally on three runs).

simonprickett:
@leslie-heinzen as discussed in Slack, my latest commit leaves the new query editor experience out as that's not currently the experience the intended user gets in play, but I've addressed the other points.

leslie-heinzen (approved):
LGTM! To confirm, with the changes you made the guide passes all E2E runner checks. Thanks!

leslie-heinzen:
@simonprickett You can merge this one whenever you're ready. Previous approval satisfies merge checks now. 👍🏻

PR #466, opened 2026-07-20
Compute manifest rollup stats before build-repository (RFC #14 spike)

moxious:
## SPIKE — draft for team discussion

**Stays a draft** pending approval of RFC [grafana/pathfinder-rfcs#14 — *Step-level progression tracking for learning paths*](https://github.com/grafana/pathfinder-rfcs/pull/14) and final decisions on the `stats` field shape. Do not merge; do not mark ready-for-review.

### What this is

A CI build step + local helper that computes **rollup stats** on every package `manifest.json`, run on the staged package tree immediately before each `build-repository` invocation.

- Helper: `.github/scripts/compute_manifest_stats.py` (pure Python, stdlib only — matches the existing `validate_json.py` / `validate_html.py` style; no credentials required).
- CI wiring: a new step in `deploy.yml` (`build-packages` job) and `validate-json.yml` (`validate-packages` job), each placed immediately before `build-repository`.

### Build-time-only posture (no committed stats)

**This PR commits NO `stats` into any `manifest.json`.** The helper is **additive at build time only**: it mutates the ephemeral CI checkout (and, locally, whatever tree you point it at) so that `build-repository` sees enriched manifests, but it never writes `stats` back into the committed repo. Committed stats would drift out of sync with content edits and be consistently wrong, so they are computed fresh on every build instead. Acceptance is demonstrated below via **sample outputs pasted into this PR body** and the **CI step output**, not via committed files. The net diff of this PR is exactly the script plus the two workflow edits.

### Two-PR design

This is the **compute** half of a two-PR spike:

1. **This PR (interactive-tutorials)** — computes per-guide `stats` and embeds them in each `manifest.json` at build time; rolls guide stats up onto path/journey manifests.
2. **Companion PR (grafana-pathfinder-app, built in parallel — URL not known at authoring time)** — relaxes the CLI's `build-repository` so the `stats` field on manifests is **carried through into `repository.json`**. Today `build-repository` does not emit `stats`.

**Pinned-SHA dependency:** both workflows pin the pathfinder-app CLI to `342121e6ef884afeddcc680f2a8f6ca60b687488`. Until the companion change lands **and** that pin is bumped past it, `stats` is embedded in the (build-time) `manifest.json` files but **will NOT propagate into `repository.json`** — that is expected for this spike. This PR's observable acceptance is on the `manifest.json` files themselves (as produced at build time).

### Output shape

**Guide** (`type: guide`, measured from its sibling `content.json`):

```json
"stats": { "version": 1, "steps": 17, "blocks": 25, "sections": 3 }
```

**Path / journey** (`type: path` or `journey`) — sums over `milestones`, plus milestone count and a per-milestone breakdown **keyed by milestone id (not position)**:

```json
"stats": {
  "version": 1,
  "guides": 7,
  "steps": 13,
  "blocks": 69,
  "sections": 6,
  "perMilestone": {
    "fleet-mgt-monitor-health-determine-config":       { "steps": 4, "blocks": 10, "sections": 1 },
    "fleet-mgt-monitor-health-create-token":           { "steps": 2, "blocks": 9,  "sections": 1 },
    "fleet-mgt-monitor-health-register-collector":     { "steps": 4, "blocks": 16, "sections": 1 },
    "fleet-mgt-monitor-health-check-health-status":    { "steps": 2, "blocks": 6,  "sections": 1 },
    "fleet-mgt-monitor-health-view-health-dashboards": { "steps": 1, "blocks": 13, "sections": 1 },
    "fleet-mgt-monitor-health-view-collector-logs":    { "steps": 0, "blocks": 13, "sections": 1 },
    "fleet-mgt-monitor-health-end-journey":            { "steps": 0, "blocks": 2,  "sections": 0 }
  }
}
```

Both samples above are **real output** of the helper (guide: `first-dashboard`; path: `fleet-mgt-monitor-health-lj`), captured by running it on a scratch copy — not committed to the repo.

Ordering guarantee: **all guides are measured first** (building an id → stats map), then paths/journeys roll up from that map, so a milestone's stats are always available before its rollup. This is enforced structurally by two separate passes over the loaded manifests, independent of filesystem walk order (which is itself sorted for determinism).

### Counting semantics

`steps` is an **exact port** of the authoritative runtime helper `count-guide-steps.ts` (grafana-pathfinder-app). The requested branch `backup/1370-step-weights` has since been **renamed to `feat/journey-step-weights`**; the file lives at `src/docs-retrieval/count-guide-steps.ts`. The port reproduces all 11 of the reference `count-guide-steps.test.ts` cases (verified locally, including the `assistantEnabled`-in-section case). Step block types come from `STEP_COUNTING_BLOCK_TYPES`:
`interactive`, `multistep`, `guided`, `quiz`, `terminal`, `terminal-connect`, `challenge`, `code-block`.

Matching the runtime exactly, `steps` traversal recurses **only** into `section` blocks, so:
- `conditional` branches (`whenTrue`/`whenFalse`) count **0**.
- unresolved `snippet-ref` counts **0**.
- an `assistantEnabled` step **inside a section** counts **0** (render wrapper hides it from section extraction); the same step at top level counts 1.
- `multistep` / `guided` count **1** each, regardless of their inner `steps` array length.

### Documented approximations (acceptable for a spike)

1. **`blocks` and `sections` use a fuller traversal than `steps`.** They descend through every content-block-bearing array — section `blocks` **and** conditional `whenTrue`/`whenFalse` — deliberately, so conditional content is not invisible in the size metric. Consequence: a section or step nested inside a `conditional` is counted in `blocks`/`sections` but **not** in `steps` (which mirrors the runtime). `multistep`/`guided` inner `steps` are action descriptors, not content blocks, so each such block counts as exactly one block.
2. **`blocks` counts every block node, containers included** — a `section` or `conditional` counts as one block in addition to its descendants.
3. **`sections`** counts every block of `type: section` anywhere in the tree (nested sections included).
4. **Rollup linkage is by milestone `id`**, not directory name or position (a milestone string resolves to the guide manifest whose `id` equals it). A milestone that resolves to no measured guide is skipped from `perMilestone`/sums with a warning; `guides` still reflects the full milestone count. Only `type: guide` manifests are measured, so a **nested journey** (a milestone referencing another `path`/`journey`) contributes nothing to the sums while still counting toward `guides` — a known gap flagged for the RFC; no such nesting exists in this repo today. Demonstrated: `fleet-mgt-monitor-health-lj/business-value-fleet-mgt` is a `guide` directory **not** listed in the path's `milestones` — it gets its own `stats` but is correctly **absent** from the path rollup.
5. A path/journey's own sibling `content.json` (if any) is **not** counted — a path is defined purely as the rollup of its milestones.

(The same list is in the script header.)

### Determinism

`stats` is written with fixed key order and no timestamps. The writer is **surgical**: it excises any existing `stats` member (JSON-structure-aware) and re-inserts it identically, preserving every other byte of the manifest (no whole-file reformat — important, since ~285/479 manifests use compact inline arrays that a naive `json.dumps` would churn). Re-running on unchanged content is a byte-for-byte no-op — the CI step output over the full tree:

```
$ python3 .github/scripts/compute_manifest_stats.py .
📊 Manifests processed: 479 (430 guides, 49 paths/journeys)
✅ Updated 479 manifest(s); 0 already current
$ python3 .github/scripts/compute_manifest_stats.py .        # second run
✅ Updated 0 manifest(s); 479 already current
$ python3 .github/scripts/compute_manifest_stats.py . --check
✅ All manifest stats are up to date   # exit 0
```

**Manifests processed: 479 (430 guides + 49 paths/journeys)**, 0 unresolved milestones across the repo.

### Validation

Unknown manifest fields are **not rejected** by the pinned CLI's validation (verified), so `stats` passes through `build-repository`. All 479 enriched manifests were confirmed well-formed JSON locally. Because enrichment is build-time-only, the existing `validate-packages` job runs `validate --package` on the committed (un-enriched) source first, then this step enriches the ephemeral tree before `build-repository` — so no committed file is altered and existing validation is unaffected.

### Known limitation — `guides/` tree does not carry stats (open discussion point)

`deploy.yml`'s legacy `deploy-guides` job copies manifests into the published `guides/` tree **independently** of the `build-packages` job, and the new compute step only enriches the `packages/` staging tree. So under this wiring the `guides/` tree would **not** carry stats. Left unaddressed intentionally — it's an open discussion point for the RFC (the `guides/` tree is the legacy publication path). Not fixed here.

### Security partitioning preserved

The new step in `build-packages` is pure Python/stdlib and adds no permissions; `build-packages` still holds **no `id-token`** (no credentials), unchanged. No job layout was restructured — the previous single "Build Pathfinder Repo" step was split into stage-copy / compute-stats / build-repository within the same sandboxed job.

### DIFF
--- a/.github/scripts/compute_manifest_stats.py
+++ b/.github/scripts/compute_manifest_stats.py
@@ -1,0 +1,459 @@
+#!/usr/bin/env python3
+"""
+Manifest rollup-stats computation for the GitHub Actions build pipeline.
+
+SPIKE (RFC grafana/pathfinder-rfcs#14 — "Step-level progression tracking for
+learning paths"). Computes per-guide content statistics and embeds them into
+each package manifest.json as a `stats` object, then rolls those stats up onto
+each learning path / journey manifest. Runs on the *staged* package tree just
+before the pathfinder-app CLI's `build-repository` step.
+
+Ordering: every `guide` manifest is measured first, building an id -> stats
+map, and only then are `path`/`journey` manifests rolled up from that map.
+This guarantees a path's milestone stats are available before its rollup.
+
+Output shape
+------------
+guide manifest:
+    "stats": { "version": 1, "steps": N, "blocks": N, "sections": N }
+
+path / journey manifest (sums over its `milestones`, keyed by milestone id):
+    "stats": {
+      "version": 1,
+      "guides": N,            # number of milestones listed on the path
+      "steps": N,             # sum of milestone steps
+      "blocks": N,            # sum of milestone blocks
+      "sections": N,          # sum of milestone sections
+      "perMilestone": {       # keyed by milestone *id*, not position
+        "<milestone-id>": { "steps": N, "blocks": N, "sections": N },
+        ...
+      }
+    }
+
+Counting semantics
+------------------
+`steps` is an exact port of the authoritative runtime helper
+`count-guide-steps.ts` (grafana-pathfinder-app, branch feat/journey-step-weights,
+src/docs-retrieval/count-guide-steps.ts; originally requested as
+backup/1370-step-weights, which has since been renamed). A block counts as one
+step iff its `type` is in STEP_COUNTING_BLOCK_TYPES. The traversal recurses
+ONLY into `section` blocks. Consequently, and matching the runtime:
+  * `conditional` branches (whenTrue/whenFalse) count 0 — not recursed.
+  * `snippet-ref` counts 0 — unresolved, not a step type.
+  * an `assistantEnabled` step *inside a section* counts 0 (the render wrapper
+    hides it from section extraction); the same step at top level counts 1.
+  * `multistep` / `guided` count 1 each regardless of their inner `steps` array.
+
+DOCUMENTED APPROXIMATIONS (acceptable for this spike; see PR body):
+  1. `blocks` and `sections` use a DIFFERENT, fuller traversal than `steps`:
+     they descend through every array that holds content blocks — `blocks`
+     (sections) AND `conditional` branches `whenTrue` / `whenFalse`. This is a
+     deliberate choice: `steps` mirrors the runtime exactly, while `blocks`/
+     `sections` describe total authored content (so conditional content is not
+     invisible). It means a section or step nested inside a conditional is
+     included in `blocks`/`sections` but NOT in `steps`. `multistep`/`guided`
+     inner `steps` are action descriptors, not content blocks, so they are
+     never counted as blocks (each such block counts as exactly one block).
+  2. `blocks` counts EVERY block node, container blocks included: a `section`
+     or `conditional` counts as one block in addition to its descendants.
+  3. `sections` counts every block whose `type` is `section`, anywhere in the
+     tree (nested sections included), via the same fuller traversal as (1).
+  4. Rollup linkage is by milestone `id` (a guide manifest whose `id` equals
+     the milestone string), NOT by directory name or position. A milestone that
+     resolves to no measured guide is skipped from `perMilestone` and the sums,
+     and a warning is emitted; `guides` still reflects the full milestone count.
+  5. A path/journey's own sibling content.json (if any) is NOT counted — a path
+     is defined purely as the rollup of its milestones.
+
+Determinism: `stats` is written with a fixed key order and no timestamps.
+Re-running on unchanged content is a byte-for-byte no-op: an existing `stats`
+member is excised (JSON-structure-aware) and re-inserted identically, and all
+other bytes of the manifest are preserved exactly (no whole-file reformat).
+
+Usage:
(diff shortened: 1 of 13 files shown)


moxious:
Closing this — we're taking a different direction.

An architecture session on 2026-08-19 settled the completion denominator as a total block count, which changes the counting semantics this script implements. More importantly, it made the case for a single canonical implementation: this repo's deploy workflow already builds and invokes the pathfinder-app CLI, so a standalone Python port of the counting logic means two implementations of one traversal in two languages, kept in step by hand.

The computation is moving into `grafana-pathfinder-app` as a low-tier TypeScript module with a thin CLI command over it. Once that lands, the change needed here reduces to invoking that command in the build pipeline — a different enough shape that it belongs in a fresh PR rather than a rewrite of this one.

Nothing here is lost: the counting semantics and the documented edge cases in `compute_manifest_stats.py` were useful input to the new implementation.

PR #468, opened 2026-07-21
Remove targeting for learning path milestones

jdbaldry:
They don't need targeting because it causes those milestones to get recommended in Pathfinder separately to the learning path.

Signed-off-by: Jack Baldry <jack.baldry@grafana.com>


moxious (approved):

bonnywelsford-source (approved):
LGTM (removing milestone targeting so these steps aren’t recommended outside the learning path).

PR #470, opened 2026-07-22
Updating SM Overview Tutorials

TheComeUpCode:
Remove alerts section since its legacy feature, fixing a few new issues that popped up


moxious (changes requested):
Did something change about the SM interface?  I don't think you changed this part of the guide, but I get stuck at this point

Image

PR #474, opened 2026-07-23
Hide Grafana Play on find-k6-limits path

bonnywelsford-source:
## Summary
- Set `show_play: false` on `find-k6-limits-lj/website.yaml` so sync hides **Try in Grafana Play** for this k6 path (k6 is not available in Play).
- Matches existing k6 paths such as `run-first-k6-test` and `k6-soak-testing`.

## Test plan
- [ ] Confirm `show_play: false` is present in `find-k6-limits-lj/website.yaml`
- [ ] After website sync, verify `/docs/learning-paths/find-k6-limits/` CTA has no Play button


tacole02 (approved):
Thanks, @bonnywelsford-source !

PR #487, opened 2026-07-29
Add read-a-trace interactive learning path

knylander-grafana:
Adds the `read-a-trace` interactive learning path: a queryless, hands-on walkthrough that teaches users to read a distributed trace on the public play.grafana.org QuickPizza demo, with no account or setup.

## What it teaches
Trace *anatomy* — how to read a trace someone hands you: the span waterfall shape, parent/child nesting, span durations, service and operation names, and span attributes. It deliberately does not cover the Traces Drilldown investigation workflow or writing TraceQL; those are linked as follow-on paths.

## Milestones
1. **What you'll do** — orientation and the trace/span vocabulary
2. **Open a real trace** — runs a ready-made TraceQL query in Explore and opens a trace
3. **Read the span waterfall** — panel = one trace, nesting = call chain, bar length = duration
4. **Services, operations, and the database fan-out** — reading row names and expanding children
5. **Open a span for its details** — reading attributes
6. **Healthy vs. failed requests** (optional) — an error query showing how a failed trace's shape alone tells the story
7. **What the waterfall told you** — recap and follow-on paths (Drilldown, Alloy, visualization, TraceQL cookbook)

## Validation
- **Deep links use the shipped `grafanacloud-traces` UID.** Verified anonymously on play.grafana.org: both the healthy (`duration > 100ms`) and error (`status = error`, `now-3h`) queries load pre-filled and return traces.
- **Selectors resolve in the live trace view:** panel header (1 match), indent guide and span bar pinned with `:nth-match(1)`.
- **SME has confirmed** both TraceQL queries are correct.
- Passes `lint` and `check`; `attack` findings addressed (removed auto-completing `on-page` objectives from navigate steps, added trace-open recovery guidance and reload-query notes, softened optional Critical path / Errors chip wording).

## Publishing
On merge, `scripts/interactive-learning/sync` generates the website page at `/docs/learning-paths/read-a-trace/` (slug = package id minus `-lj`). This matches the `{{< learning-path path="/docs/learning-paths/read-a-trace/" >}}` embed in the intro-to-tracing journey (slide 15).

## Also in this PR
- Adds a CODEOWNERS entry for the new path.


knylander-grafana:
This interactive tutorial is set to use play.grafana.org with the default data source for play -- the premise is that a user doesn't need to have an account or any data. What does that mean for the visibility of this Pathfinderized thingy? Will it only show up on play?

melgl (dismissed):
Looks good! 
- content is accurate
- user flow is sensible and easy to follow
- confirmed all links on last resource slide successfully resolve

read-a-trace-lj/content.json
```diff
@@ -0,0 +1,10 @@
...
+      "type": "markdown",
+      "content": "Welcome to the Grafana Labs learning path that teaches you to read a distributed trace, hands-on, on the public play.grafana.org demo.\n\nA trace records the whole journey of one request as it moves through the services of a distributed system. Reading one tells you what each step did and where the request spent its time. That skill is the same wherever you open a trace, so learning it once carries everywhere.\n\n## Here's what to expect\n\nWhen you complete this path, you'll be able to:\n\n- Open a real trace and take in the shape of its span waterfall at a glance\n- Follow parent and child nesting to see which operation triggered which\n- Read span durations and see where a request spent its time\n- Read a service and operation name, and open a span for its attributes\n\n## Before you begin\n\nYou don't need a Grafana Cloud account or any setup. Everything runs on the public play.grafana.org demo, which streams live trace data you can open right away.\n\nThis path teaches trace **anatomy**: how to read a trace someone hands you. It doesn't teach the Traces Drilldown investigation workflow or how to write a query; those come later."
```
tiffanyfay: > Follow parent and child nesting to see which operation triggered which

What does this mean? Parents trigger children?

tiffanyfay: > Read span durations

Span durations are how long a span takes, so read feels weird

tiffanyfay: > You don't need a Grafana Cloud account or any setup. Everything runs on the public play.grafana.org demo, which streams live trace data you can open right away.

This feels weird because if they're on here they're already on play.

tiffanyfay: > how to read a trace someone hands you

but what if you just came across the trace yourself?

read-a-trace-lj/welcome/content.json
```diff
@@ -0,0 +1,18 @@
+{
+  "id": "read-a-trace-welcome",
+  "title": "What you'll do",
+  "blocks": [
+    {
+      "type": "markdown",
+      "content": "You're about to open a real trace on the public play.grafana.org demo and read it top-down. The demo streams live traces you can open right away."
```
tiffanyfay: You're already here on play though

read-a-trace-lj/open-a-trace/content.json
```diff
@@ -0,0 +1,34 @@
+{
+  "id": "read-a-trace-open-a-trace",
+  "title": "Open a real trace",
+  "blocks": [
+    {
+      "type": "markdown",
+      "content": "To save you hunting for a good example, we've filled in a ready-made query that lands you on a set of traces worth reading. That query is written in TraceQL, Tempo's trace query language. You don't need to read or write it here; the step below opens Explore and runs it for you. You'll see where TraceQL fits at the end of this path.\n\nRun the ready-made query below to load a set of example traces."
```
tiffanyfay: I'd add a link here or somewhere on this page to the TraceQL docs

read-a-trace-lj/read-the-waterfall/content.json
```diff
@@ -0,0 +1,55 @@
...
+        }
+      ]
+    },
+    {
+      "type": "markdown",
+      "content": "That's the trace's shape and timing in three reads: the panel is one trace, the indentation shows what called what, and the bars show how long each step took."
+    },
+    {
+      "type": "markdown",
+      "content": "Now, find the longest bar in the trace. That single span is where the request spent most of its time. Its name tells you what it was, for example a database query, an API call, or a compute step. Reading a trace top-down like this is how you spot where the time went without checking every row."
```
tiffanyfay: > Now, find the longest bar in the trace. 

This might be a little confusing since the longest bar is the first one as it's always the total

tiffanyfay: > Its name tells you what it was, for example a database query, an API call, or a compute step.

Is this the span name? Because there isn't a rule to the naming. The operation may tell this though.

Image

Image

read-a-trace-lj/read-the-waterfall/content.json
```diff
@@ -0,0 +1,55 @@
...
+    },
+    {
+      "type": "markdown",
+      "content": "Optional: if your view shows a **Critical path** option, select it to shade the hops that actually drove the end-to-end time, so on a busier trace you can pick out the slow path at a glance."
```
tiffanyfay: I'd say critical path button and placement of it -- unsure what is the best wording choice of whether it is under filters or above service and operation text

Also for me, it lit up every single span except one, so that alone doesn't help in some cases. What they need to find is a span with the longest black line/critical path from my understanding.

tiffanyfay: Image

read-a-trace-lj/services-and-fan-out/content.json
```diff
@@ -0,0 +1,22 @@
+{
+  "id": "read-a-trace-services-and-fan-out",
+  "title": "Services, operations, and the database fan-out",
```
tiffanyfay, line 3: I feel like this section would benefit from images

read-a-trace-lj/span-attributes/content.json
```diff
@@ -0,0 +1,18 @@
...
+      "type": "markdown",
+      "content": "With your trace still open, you can dig into any single step. The waterfall tells you the shape and the timing; for the details of one step, open the span itself."
+    },
+    {
+      "type": "markdown",
+      "content": "Select any span row to open its details. Its **attributes** appear right below the row: the key-value facts about that step, for example the request's URL, the status code it returned, or the database statement it ran."
+    },
+    {
+      "type": "markdown",
+      "content": "A span's name, its duration, and a couple of attributes are usually enough to tell what that step did and whether it's the problem. That's the whole loop: read the shape top-down, find where the time went, then open a span to confirm what it was doing."
```
tiffanyfay, line 32: I think also mentioning if it's on the critical path. 

Maybe worth also mentioning spans with errors? Edit: I see you mention this next section.

read-a-trace-lj/healthy-vs-failed/content.json
```diff
@@ -0,0 +1,38 @@
...
+        }
+      ]
+    },
+    {
+      "type": "markdown",
+      "content": "Now, open one of the error traces by clicking its trace ID."
+    },
+    {
+      "type": "markdown",
+      "content": "**If the list is empty, that's expected.** These errors come in bursts, so a quiet moment can return nothing. Run the query again in a few minutes and one will appear. If you change the query by mistake, run the query step again to reload it."
```
tiffanyfay: I wouldn't say it's expected. I would say it's not an issue. I think the default time is 1h and I see at least 15 traces with errors. And I would instead just suggest they increase the time period vs waiting around for a few minutes. I can't recall if some of this is relevant to the previous time you mention this since you did suggest extending the window.

read-a-trace-lj/healthy-vs-failed/content.json
```diff
@@ -0,0 +1,38 @@
...
+    },
+    {
+      "type": "markdown",
+      "content": "Notice how different it looks. A successful order runs the whole pipeline: many spans, the deep recommendations → catalog → SELECT fan-out. A failed one is much shorter: it short-circuits early, stopping at the step that failed, so it never reaches the rest of the pipeline. Whatever the exact counts, the failed trace is visibly smaller, and that shape alone tells you it stopped early, before you read a single attribute.\n\nThis is also where you meet **status codes** (a healthy request returns a success code, a failed one an error code) and, where your view offers one, an **Errors** filter for finding them. The lesson: useful traces aren't always big waterfalls. A small, short trace can tell you exactly where a request gave up."
```
tiffanyfay: I would tell people where to look for more info -- sometimes there's error messages. For these ones you can look at the span attributes for http and the status code.

read-a-trace-lj/end-journey/content.json
```diff
@@ -0,0 +1,26 @@
...
+  "id": "read-a-trace-end-journey",
+  "title": "What the waterfall told you",
+  "blocks": [
+    {
+      "type": "markdown",
+      "content": "You just opened a real trace and read it on your own, with no setup and no query you had to write. That's the core skill, and it transfers to any trace on any surface."
+    },
+    {
+      "type": "markdown",
+      "content": "Here's what you pulled out of the span waterfall:\n\n- Took in the trace's shape at a glance.\n- Followed the parent and child nesting to see which call ran inside which.\n- Read the durations and found where the request spent its time.\n- Learned to open a span and read its attributes to tell what that step did."
```
tiffanyfay, line 11: IDK where exactly, but I feel maybe a little explanation of what span attributes are and resource attributes are if I didn't miss it.

read-a-trace-lj/end-journey/content.json
```diff
@@ -0,0 +1,26 @@
+{
+  "id": "read-a-trace-end-journey",
+  "title": "What the waterfall told you",
+  "blocks": [
+    {
+      "type": "markdown",
+      "content": "You just opened a real trace and read it on your own, with no setup and no query you had to write. That's the core skill, and it transfers to any trace on any surface."
```
tiffanyfay: > and no query you had to write

feels kinda awkward because the guide dumps in the query vs not needing queries at all -- which is what would be the case using search. Is there a reason to even be using traceql here? 

the last two parts of this sentence feel weird as well with the next line about core skill since the core skill is looking through a trace and understanding/reading it. Or is this just me not getting grammar?

read-a-trace-lj/end-journey/content.json
```diff
@@ -0,0 +1,26 @@
...
+    },
+    {
+      "type": "markdown",
+      "content": "Here's what you pulled out of the span waterfall:\n\n- Took in the trace's shape at a glance.\n- Followed the parent and child nesting to see which call ran inside which.\n- Read the durations and found where the request spent its time.\n- Learned to open a span and read its attributes to tell what that step did."
+    },
+    {
+      "type": "markdown",
+      "content": "Two doors open from here:\n\n- **A query language for finding traces on purpose.** You ran a ready-made TraceQL query to land on your practice traces. When your questions get specific, like \"failed checkout requests slower than two seconds,\" TraceQL is how you pin them down.\n- **Connecting a trace to your other signals.** A trace is even more useful next to the logs and metrics from the same request. That correlation is powerful, but it isn't automatic; it needs setup."
```
tiffanyfay: > That correlation is powerful, but it isn't automatic; it needs setup.

doesn't exactly feel like _you_

tiffanyfay (commented):
Great work! Left a bunch of comments.

read-a-trace-lj/end-journey/content.json
knylander-grafana: good catch. I'll ulpdate this.

read-a-trace-lj/content.json
knylander-grafana: Good point. This phrasing should be removed. I found out it can be possible to install the play (with quick pizza) data source in your own stack. In theory, that would let this interactive tutorial be used any where, not just on play or learn.

read-a-trace-lj/welcome/content.json
knylander-grafana: Good catch. I'll fix this.

knylander-grafana:
> This interactive tutorial is set to use play.grafana.org with the default data source for play -- the premise is that a user doesn't need to have an account or any data. What does that mean for the visibility of this Pathfinderized thingy? Will it only show up on play?

This interactive tutorial is currently set to only show up on play.grafana.org. For a later iteration, we can change that so it can show up on any stack--with the understanding that the user might need to install the quickpizza demo (play data source) data source to enable the queries and interactive elements.

read-a-trace-lj/services-and-fan-out/content.json
knylander-grafana: We could do images. I'll run through it. might also benefit from better selectors, too.

knylander-grafana:
> Great work! Left a bunch of comments.

Thank you! Your comments are very helpful. I've updated to meet most of them. Screenshots might get added in a later iteration, along with widening the scope of where people can seen this interactive tutorial. However, it has additional requirements that I have to research (asking for the inclusion of a specific data source, for exapmle).

read-a-trace-lj/read-the-waterfall/content.json
```diff
@@ -0,0 +1,55 @@
...
+          "action": "highlight",
+          "reftarget": "[data-testid=\"data-testid Panel header Trace\"]",
+          "content": "This whole panel is one trace. Start at the root: the top row is the request itself, the span that everything else happens inside. In this demo it's the quickpizza-public-api span for the pizza-order request.",
+          "doIt": false,
+          "requirements": ["on-page:/explore", "exists-reftarget"]
+        },
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "[data-testid=\"SpanTreeOffset--indentGuide\"]:nth-match(1)",
```
tacole02: In Block Editor, Show me/Do it failed on this indent-guide highlight (`SpanTreeOffset--indentGuide`:nth-match(1)), and Assistant could not recover, so I could not finish the span-bar step after it. Can you re-capture a stable selector against an open QuickPizza trace (or drop the highlight if the UI does not expose a reliable target)?

read-a-trace-lj/open-a-trace/content.json
```diff
@@ -0,0 +1,34 @@
...
+    },
+    {
+      "type": "section",
+      "blocks": [
+        {
+          "type": "interactive",
+          "action": "navigate",
+          "reftarget": "/explore?schemaVersion=1&panes=%7B%22trc%22%3A%7B%22datasource%22%3A%22grafanacloud-traces%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22datasource%22%3A%7B%22type%22%3A%22tempo%22%2C%22uid%22%3A%22grafanacloud-traces%22%7D%2C%22queryType%22%3A%22traceql%22%2C%22query%22%3A%22%7B%20rootServiceName%20%3D%20%5C%22quickpizza-public-api%5C%22%20%26%26%20rootName%20%3D%20%5C%22POST%20%2Fapi%2Fpizza%5C%22%20%26%26%20duration%20%3E%20100ms%20%7D%22%2C%22limit%22%3A20%2C%22tableType%22%3A%22traces%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D&orgId=1",
+          "content": "This step opens Explore and runs the ready-made query for you, returning a list of candidate traces that are all the same kind of request. Wait for the results table to load.",
```
tacole02: This navigate step works, but the copy reads like system narration ("This step opens…") instead of telling the learner what to do. Same pattern on healthy-vs-failed. Both milestones are also mostly markdown around a single interactive. Prefer action-focused step text, and consider whether more of the click-trace / compare beats can be interactive.

read-a-trace-lj/manifest.json
```diff
@@ -0,0 +1,41 @@
...
+        }
+      ]
+    }
+  },
+  "testEnvironment": {
+    "tier": "cloud",
+    "instance": "play.grafana.org"
+  },
+  "milestones": [
+    "read-a-trace-welcome",
```
tacole02: `read-a-trace-welcome` looks like framing (orientation / why). Framing packages usually stay on disk with website.yaml but stay out of Pathfinder `milestones`, with the first hands-on using `depends: []`. If welcome is meant as a real path step learners must complete, ignore this; otherwise drop it from the array and update `open-a-trace` depends.

read-a-trace-lj/read-the-waterfall/content.json
```diff
@@ -0,0 +1,55 @@
...
+      "type": "markdown",
+      "content": "That's the trace's shape and timing in three reads: the panel is one trace, the indentation shows what called what, and the bars show how long each step took."
+    },
+    {
+      "type": "markdown",
+      "content": "The top bar is always the whole request, so skip it and find the longest bar below it. That span is where the request spent most of its time. Its service and operation names tell you what it was, for example a database query, an API call, or a compute step. Reading a trace top-down like this is how you spot where the time went without checking every row."
+    },
+    {
+      "type": "markdown",
+      "content": "Optional: if your view offers a **Critical path** control (look near the filters, or above the service and operation names), select it to shade the hops that drove the end-to-end time. It can highlight most of the spans, so don't just look for highlighted rows. Look for the span with the longest critical-path segment, the solid black line, because that's where the slow path actually spent its time."
```
tacole02: Docs describe the critical path as a darker segment on the duration bar, not a "solid black line" (theme-dependent). Worth matching that wording so learners are not hunting for a black line.

read-a-trace-lj/read-the-waterfall/content.json
```diff
@@ -0,0 +1,55 @@
...
+        }
+      ]
+    },
+    {
+      "type": "markdown",
+      "content": "That's the trace's shape and timing in three reads: the panel is one trace, the indentation shows what called what, and the bars show how long each step took."
+    },
+    {
+      "type": "markdown",
+      "content": "The top bar is always the whole request, so skip it and find the longest bar below it. That span is where the request spent most of its time. Its service and operation names tell you what it was, for example a database query, an API call, or a compute step. Reading a trace top-down like this is how you spot where the time went without checking every row."
```
tacole02: "Always the whole request" and treating the longest child bar as where time went are a bit absolute. Soften those, and point people at critical path when you care about end-to-end time.

read-a-trace-lj/span-attributes/content.json
```diff
@@ -0,0 +1,22 @@
...
+  "id": "read-a-trace-span-attributes",
+  "title": "Open a span for its details",
+  "blocks": [
+    {
+      "type": "markdown",
+      "content": "With your trace still open, you can dig into any single step. The waterfall tells you the shape and the timing; for the details of one step, open the span itself."
+    },
+    {
+      "type": "markdown",
+      "content": "Select any span row to open its details. Its **attributes** appear right below the row. Attributes are the key-value facts attached to a trace. **Span attributes** describe that one step, for example the request's URL, the status code it returned, or the database statement it ran. **Resource attributes** describe where the span came from, for example the service name, host, or environment that produced it."
```
tacole02: Attributes are attached to spans (and resources/events), not to the trace as a whole. Suggest "key-value facts on a span" here, then keep your span vs resource split.

read-a-trace-lj/healthy-vs-failed/content.json
```diff
@@ -0,0 +1,38 @@
...
+        }
+      ]
+    },
+    {
+      "type": "markdown",
+      "content": "Now, open one of the error traces by clicking its trace ID."
+    },
+    {
+      "type": "markdown",
+      "content": "**If the list is empty, that's not an issue.** Widen the time range at the top of Explore and run the query again. Error traces are common, so a larger window will surface some. If you change the query by mistake, run the query step again to reload it."
```
tacole02: "Error traces are common" is not something we can source. Keep the widen-the-time-range recovery; drop "common."

read-a-trace-lj/healthy-vs-failed/content.json
```diff
@@ -0,0 +1,38 @@
...
+    },
+    {
+      "type": "markdown",
+      "content": "Notice how different it looks. A successful order runs the whole pipeline: many spans, the deep recommendations → catalog → SELECT fan-out. A failed one is much shorter: it short-circuits early, stopping at the step that failed, so it never reaches the rest of the pipeline. Whatever the exact counts, the failed trace is visibly smaller, and that shape alone tells you it stopped early, before you read a single attribute.\n\nThis is also where you meet **status codes** (a healthy request returns a success code, a failed one an error code) and, where your view offers one, an **Errors** filter for finding them. To see why a request failed, open the span that failed and read its attributes: the HTTP attributes and status code, for example http.status_code, often carry an error message that names the cause. The lesson: useful traces aren't always big waterfalls. A small, short trace can tell you exactly where a request gave up."
```
tacole02: Two accuracy nits in this block: `http.status_code` is the numeric status (cause text usually lives in span events/exceptions), and failed traces do not always short-circuit before the rest of the pipeline. Soften the absolutes and send people to events for the error message.

tacole02 (commented):
Thanks for the PR. I smoke-tested this on learn.grafana.net shared in Block Editor.

The path walks through reading a QuickPizza span waterfall. The navigate milestones ran, but the indent-guide highlight in read-the-waterfall failed and blocked the next step. I left a few notes on the diff.

read-a-trace-lj/manifest.json
```diff
@@ -0,0 +1,47 @@
...
+          "urlPrefix": "/explore"
+        },
+        {
+          "or": [
+            {
+              "source": "play.grafana.org"
+            },
+            {
+              "source": "learn.grafana.net"
+            }
```
tacole02, line 25: If you want this path recommended only on Grafana Play (not learn.grafana.net / other stacks), drop the `or` and keep a single `source` in the `and` with `urlPrefix`:

```json
"targeting": {
  "match": {
    "and": [
      { "urlPrefix": "/explore" },
      { "source": "play.grafana.org" }
    ]
  }
}
```

`source` is what the recommender uses for instance routing (`testEnvironment.instance` below is only for testing). Right now the `or` with `learn.grafana.net` will also recommend it there.

tacole02:
> This interactive tutorial is set to use play.grafana.org with the default data source for play -- the premise is that a user doesn't need to have an account or any data. What does that mean for the visibility of this Pathfinderized thingy? Will it only show up on play?

Good question, @knylander-grafana, and sorry I missed it up-thread . I added an inline comment to the root manifest file explaining how to only surface this guide in the recommender for Play.

Other than that, this learning path looks great! I'll preemptively approve, and you can feel free to merge after adjusting the recommender settings. Great work!

tacole02 (dismissed):

read-a-trace-lj/content.json
knylander-grafana: I've rewritten some of the text to make it read a bit better. What do you think?

knylander-grafana: I redid the section, so hopefully this feels better now.

knylander-grafana: This phrasing is removed now. Agreed that it was confusing.

read-a-trace-lj/open-a-trace/content.json
knylander-grafana: Done.

read-a-trace-lj/read-the-waterfall/content.json
knylander-grafana: Clarified. Thank you for catching this.

knylander-grafana: Updated the text to separate the two ro labels and attribute the "what did this step do" so it's correct now. Thank you for pointing this out.

read-a-trace-lj/read-the-waterfall/content.json
knylander-grafana: Taylor also found issues in this section, so I've removed the interaction on this slide. There were too many issues with it. I've changed the wording to point out that a critical path might not be present.

read-a-trace-lj/end-journey/content.json
knylander-grafana: It should be updated now and read better.

knylander-grafana:
> > This interactive tutorial is set to use play.grafana.org with the default data source for play -- the premise is that a user doesn't need to have an account or any data. What does that mean for the visibility of this Pathfinderized thingy? Will it only show up on play?
> 
> Good question, @knylander-grafana, and sorry I missed it up-thread . I added an inline comment to the root manifest file explaining how to only surface this guide in the recommender for Play.

Since the data source used exists in both learn and play, I've updated the PR to use both.  Thank you for the example!
 
> Other than that, this learning path looks great! I'll preemptively approve, and you can feel free to merge after adjusting the recommender settings. Great work!

Thank you! Unfortunately, I can't get this to merge because I don't have merge rights, even with an approved PR. Would you be willing to merge it?

GrafanaWriter (approved):

tacole02 (approved):
Merging on behalf of @knylander-grafana 😄

PR #489, opened 2026-07-29
Add getting-started-with-Explore guide

Jayclifford345:
## Summary
- New standalone interactive guide, `explore-getting-started`, targeted at `/explore` on Grafana Cloud
- Tours the shared Explore interface (toolbar, query row, results controls), then walks through a full write-a-query-to-navigate-the-results flow for **metrics** (PromQL), **logs** (LogQL), and **traces** (TraceQL)
- First section installs the shared QuickPizza demo data sources (`grafanacloud-play-prom`/`-logs`/`-traces`) via the existing Demo Data Dashboards flow, so every later step and example query is identical for any Grafana Cloud user — auto-skips if already installed
- Recommends `saved-queries-explore` as a natural follow-up

## Why
Requested as an extensive, ground-up "getting started with Explore" guide covering general UI orientation plus a deep, hands-on flow for each signal type — standardized on shared demo data instead of guessing at whatever data sources a given org happens to have, so the guide behaves the same for everyone.

## How this was verified
- Selectors captured live against a real Grafana Cloud Explore session (Playwright), not guessed
- `pathfinder-cli validate --package explore-getting-started` passes clean
- Example queries (`quickpizza_server_http_request_duration_seconds_count`, `service_name="quickpizza-catalog"`, `quickpizza-public-api`) confirmed live against the actual QuickPizza demo dataset

## Test plan
- [ ] Open in the Pathfinder Block Editor against a live Grafana Cloud stack and click through end-to-end
- [ ] Confirm the two Loki-specific selectors inferred by naming-pattern (`Panel header Logs volume`, `loki options`) resolve correctly — flagged with `skippable`/`hint` in case they don't
- [ ] Confirm `install-demo-datasources` section correctly auto-skips on a stack that already has the QuickPizza demo installed, and correctly installs it on one that doesn't

Marked as draft pending that live click-through pass.

### DIFF
--- a/explore-getting-started/content.json
+++ b/explore-getting-started/content.json
@@ -317,8 +317,8 @@
         },
         {
           "type": "interactive",
-          "action": "button",
-          "reftarget": "Metrics browser",
+          "action": "highlight",
+          "reftarget": "button[data-testid='data-testid open metrics browser']",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "doIt": false,
           "skippable": true,
@@ -439,8 +439,8 @@
         },
         {
           "type": "interactive",
-          "action": "button",
-          "reftarget": "Kick start your query",
+          "action": "highlight",
+          "reftarget": "button[data-testid='data-testid Query patterns']",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-logs", "exists-reftarget"],
           "doIt": false,
           "tooltip": "It inserts a ready-made LogQL pattern instead of starting from a blank query.",
@@ -448,8 +448,8 @@
         },
         {
           "type": "interactive",
-          "action": "button",
-          "reftarget": "Label browser",
+          "action": "highlight",
+          "reftarget": "button[data-testid='label-browser-button']",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-logs", "exists-reftarget"],
           "doIt": false,
           "skippable": true,
--- a/explore-getting-started/content.json
+++ b/explore-getting-started/content.json
@@ -59,7 +59,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "div[data-testid='data-testid Data source picker select container']",
+          "reftarget": "grafana:components.DataSourcePicker.container",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "doIt": false,
           "tooltip": "Every query starts here — Prometheus, Loki, Tempo, or anything else you've connected.",
@@ -68,7 +68,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid explore-toolbar-go-queryless-button']",
+          "reftarget": "grafana:pages.Explore.toolbar.goQueryless",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "doIt": false,
           "skippable": true,
@@ -79,7 +79,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid explore-toolbar-split-button']",
+          "reftarget": "grafana:pages.Explore.toolbar.split",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "content": "Click **Split** to open a second, independent query pane."
         },
@@ -98,7 +98,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid explore-toolbar-add-dropdown-button']",
+          "reftarget": "grafana:pages.Explore.toolbar.addTo",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "doIt": false,
           "tooltip": "Use it to open another data source in a new tab without losing your current query.",
@@ -107,7 +107,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid TimePicker Open Button']",
+          "reftarget": "grafana:components.TimePicker.openButton",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "doIt": false,
           "tooltip": "The arrows on either side step that same range backward or forward, and the magnifying glass zooms out.",
@@ -116,14 +116,14 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid RefreshPicker run button']",
+          "reftarget": "grafana:components.RefreshPicker.runButtonV2",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "content": "Click **Run query** to execute whatever is in the query editor."
         },
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid RefreshPicker interval button']",
+          "reftarget": "grafana:components.RefreshPicker.intervalButtonV2",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "doIt": false,
           "tooltip": "Handy for watching something unfold in near-real time without clicking Run query yourself.",
@@ -132,7 +132,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid explore-toolbar-share-button']",
+          "reftarget": "grafana:pages.Explore.toolbar.share",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "doIt": false,
           "tooltip": "Anyone who opens it lands on the same data source, query, and time range you're looking at.",
@@ -141,14 +141,14 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid explore-toolbar-content-outline-button']",
+          "reftarget": "grafana:pages.Explore.toolbar.contentOutline",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "content": "Click **Outline** to toggle the section-jump sidebar."
         },
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "div[data-testid='data-testid Query editor row']",
+          "reftarget": "grafana:components.QueryEditorRows.rows",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "doIt": false,
           "tooltip": "Later queries can reference an earlier one's results by that letter — useful for math across queries.",
@@ -157,7 +157,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid Duplicate query']",
+          "reftarget": "grafana:components.QueryEditorRow.actionButton:Duplicate query",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "doIt": false,
           "tooltip": "Its neighbors let you view data source help, save the query, hide its results, remove it, or drag it to reorder.",
@@ -166,7 +166,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid query-history-button']",
+          "reftarget": "grafana:components.QueryTab.queryHistoryButton",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "content": "Click **Query history** to see every query you've run recently."
         },
@@ -177,7 +177,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid query-history-button']",
+          "reftarget": "grafana:components.QueryTab.queryHistoryButton",
           "requirements": ["on-page:/explore", "exists-reftarget"],
           "content": "Click **Query history** again to close the panel."
         },
@@ -222,7 +222,7 @@
           "steps": [
             {
               "action": "highlight",
-              "reftarget": "input[data-testid='data-testid Select a data source']",
+              "reftarget": "grafana:components.DataSourcePicker.inputV2",
               "requirements": ["exists-reftarget"]
             },
             {
@@ -235,7 +235,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "div[data-testid='data-testid QueryEditorModeToggle'] div[data-testid='data-testid radio-button']:contains('Builder')",
+          "reftarget": "div{grafana:components.DataSource.Prometheus.queryEditor.editorToggle} div{grafana:components.RadioButton.container}:contains('Builder')",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "tooltip": "The rest of this section uses the visual query builder — if a previous session left this data source in Code mode, this switches it back.",
           "content": "Make sure the query editor is in **Builder** mode."
@@ -243,7 +243,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid Query patterns']",
+          "reftarget": "grafana:components.QueryBuilder.queryPatterns",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "doIt": false,
           "tooltip": "A quick way to insert a common query shape — rate, histogram, or similar — instead of starting from a blank editor.",
@@ -252,7 +252,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "input[data-testid='data-testid metric select']",
+          "reftarget": "grafana:components.DataSource.Prometheus.queryEditor.builder.metricSelect",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "doIt": false,
           "tooltip": "Type a few letters and it fuzzy-searches every metric name your Prometheus data source knows about.",
@@ -288,7 +288,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "div[data-testid='data-testid prometheus explain switch wrapper'] div[data-testid='data-testid Switch container']",
+          "reftarget": "div{grafana:components.DataSource.Prometheus.queryEditor.explain} div{grafana:components.Switch.container}",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "content": "Turn on **Explain** to see your query broken down in plain English."
         },
@@ -304,21 +304,21 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "div[data-testid='data-testid prometheus explain switch wrapper'] div[data-testid='data-testid Switch container']",
+          "reftarget": "div{grafana:components.DataSource.Prometheus.queryEditor.explain} div{grafana:components.Switch.container}",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "content": "Turn **Explain** back off."
         },
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "div[data-testid='data-testid QueryEditorModeToggle'] div[data-testid='data-testid radio-button']:contains('Code')",
+          "reftarget": "div{grafana:components.DataSource.Prometheus.queryEditor.editorToggle} div{grafana:components.RadioButton.container}:contains('Code')",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "content": "Switch to **Code** mode to write raw PromQL directly."
         },
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid open metrics browser']",
+          "reftarget": "grafana:components.DataSource.Prometheus.queryEditor.code.metricsBrowser.openButton",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "doIt": false,
           "skippable": true,
@@ -328,7 +328,7 @@
         {
           "type": "interactive",
           "action": "formfill",
-          "reftarget": "[data-testid='data-testid prometheus query field'] textarea.inputarea",
+          "reftarget": "{grafana:components.DataSource.Prometheus.queryEditor.code.queryField} textarea.inputarea",
           "targetvalue": "quickpizza_server_http_request_duration_seconds_count",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "tooltip": "It's a counter from the shared demo app — one of the same metrics every Grafana Cloud user just installed.",
@@ -337,7 +337,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid RefreshPicker run button']",
+          "reftarget": "grafana:components.RefreshPicker.runButtonV2",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "content": "Click **Run query**."
         },
@@ -348,7 +348,7 @@
         {
           "type": "interactive",
           "action": "formfill",
-          "reftarget": "[data-testid='data-testid prometheus query field'] textarea.inputarea",
+          "reftarget": "{grafana:components.DataSource.Prometheus.queryEditor.code.queryField} textarea.inputarea",
           "targetvalue": "rate(quickpizza_server_http_request_duration_seconds_count[5m])",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "content": "Replace the query with `rate(quickpizza_server_http_request_duration_seconds_count[5m])`."
@@ -356,14 +356,14 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid RefreshPicker run button']",
+          "reftarget": "grafana:components.RefreshPicker.runButtonV2",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "content": "Click **Run query** again."
         },
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "section[data-testid='data-testid Panel header Graph']",
+          "reftarget": "grafana:components.Panels.Panel.title:Graph",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "doIt": false,
           "tooltip": "Lines suit trends over time; bars and points can make sparse or discrete values easier to read.",
@@ -372,7 +372,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "div[data-testid='data-testid viz-layout-legend']",
+          "reftarget": "grafana:components.VizLegend.legend",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "doIt": false,
           "tooltip": "Click one to isolate it, or shift-click to build up a custom comparison of just the series you want.",
@@ -381,7 +381,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "section[data-testid='data-testid Panel header Raw']",
+          "reftarget": "grafana:components.Panels.Panel.title:Raw",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "doIt": false,
           "skippable": true,
@@ -391,14 +391,14 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "div[data-testid='data-testid prometheus options'] button",
+          "reftarget": "div{grafana:components.DataSource.Prometheus.queryEditor.options} button",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "content": "Expand **Options** to see legend, format, and query-type settings."
         },
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "div[data-testid='data-testid prometheus type']",
+          "reftarget": "grafana:components.DataSource.Prometheus.queryEditor.type",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
           "doIt": false,
           "tooltip": "Instant returns one value per series for the end of the time range — useful for current-state checks like this one.",
@@ -427,7 +427,7 @@
           "steps": [
             {
               "action": "highlight",
-              "reftarget": "input[data-testid='data-testid Select a data source']",
+              "reftarget": "grafana:components.DataSourcePicker.inputV2",
               "requirements": ["exists-reftarget"]
             },
             {
@@ -440,7 +440,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid Query patterns']",
+          "reftarget": "grafana:components.QueryBuilder.queryPatterns",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-logs", "exists-reftarget"],
           "doIt": false,
           "tooltip": "It inserts a ready-made LogQL pattern instead of starting from a blank query.",
@@ -459,7 +459,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "div[data-testid='data-testid Select label']",
+          "reftarget": "grafana:components.QueryBuilder.labelSelect",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-logs", "exists-reftarget"],
           "content": "Click **Select label** to see the labels available on your logs."
         },
@@ -487,7 +487,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid RefreshPicker run button']",
+          "reftarget": "grafana:components.RefreshPicker.runButtonV2",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-logs", "exists-reftarget"],
           "content": "Click **Run query**."
         },
@@ -498,7 +498,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "section[data-testid='data-testid Panel header Logs volume']",
+          "reftarget": "grafana:components.Panels.Panel.title:Logs volume",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-logs", "exists-reftarget"],
           "doIt": false,
           "skippable": true,
@@ -529,14 +529,14 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid explore-toolbar-live-button']",
+          "reftarget": "grafana:pages.Explore.toolbar.live",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-logs", "exists-reftarget"],
           "content": "Click **Live** to start tailing new log lines as they arrive."
         },
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid='data-testid explore-toolbar-live-button']",
+          "reftarget": "grafana:pages.Explore.toolbar.live",
           "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-logs", "exists-reftarget"],
           "content": "Click it again to stop and return to a normal query."
         },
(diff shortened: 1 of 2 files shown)


Jayclifford345:
## E2E selector audit — status

Ran a full start-to-finish pass over all 68 of this guide's `reftarget` selectors against the [`@grafana/e2e-selectors`](https://github.com/grafana/grafana/tree/main/packages/grafana-e2e-selectors) package.

### Converted now (work against current releases)

Five steps moved off localized text / fragile CSS onto `data-testid` selectors that **already ship**:

| Step | Was | Now |
| --- | --- | --- |
| Metrics browser | `action: button` / `"Metrics browser"` | `button[data-testid='data-testid open metrics browser']` |
| Kick start your query (Loki) | `action: button` / `"Kick start your query"` | `button[data-testid='data-testid Query patterns']` |
| Label browser | `action: button` / `"Label browser"` | `button[data-testid='label-browser-button']` |
| PromQL query fill (×2) | `textarea.inputarea` | `[data-testid='data-testid prometheus query field'] textarea.inputarea` |

The last one also fixes a latent bug: `textarea.inputarea` is Monaco's internal class and matches *every* Monaco editor on the page — in split view the fill could land in the wrong pane. `DataSource.Prometheus.queryEditor.code.queryField` has shipped since **10.4.0** and is wired in `PromQueryCodeEditor`, so scoping to it is safe today.

### Left unchanged on purpose — waiting on grafana/grafana#129669

These controls had **no** stable test id, so I added selectors upstream in grafana/grafana#129669. The guide steps are **deliberately left on their current localized selectors** so the guide keeps working against current Grafana. Switch them over once that PR ships and this guide's minimum supported version catches up:

| Step | Current selector | Upstream selector to adopt |
| --- | --- | --- |
| Builder mode | `...QueryEditorModeToggle...radio-button']:contains('Builder')` | keep the scope, swap the `:contains()` leaf → `{grafana:…queryEditor.editorToggle} {grafana:components.RadioButton.option:builder}` |
| Code mode | `...:contains('Code')` | same, with `…option:code` |
| Tempo Search | `...radio-button']:contains('Search')` | `div[data-testid='data-testid Query editor row tempo A'] {grafana:components.RadioButton.option:traceqlSearch}` |
| Tempo Service Graph | `...:contains('Service Graph')` | same, with `…option:serviceMap` |
| Close split pane | `button[aria-label='Close split pane']:nth-match(2)` | `data-testid explore-toolbar-close-split-button` |
| Query inspector (×2 steps) | `action: button` / `"Query inspector"` | `data-testid Query inspector button` |
| Log line menu | `button[aria-label='Log menu']:nth-match(1)` | `data-testid Log line menu button` |
| Log fields search | `input[placeholder='Search fields by name']` | `data-testid Log fields search input` |
| Trace span filter | `input[placeholder='Filter by attribute or text']` | `data-testid trace-page-header-adhoc-filters-row` (scope, then target the input inside) |

> ⚠️ **Keep the scoping when you swap these.** `RadioButton.option` is keyed on the option's
> `value`, which is unique within a radio group but **not across a page** — `true`/`false` alone are
> used by around a dozen different groups. Used bare it can match the wrong control, so leave the
> surrounding `editorToggle` / Tempo-query-row scope in place and only replace the `:contains()`
> leaf. Documented on the selector itself in grafana/grafana#129669.

**The four `:contains()` steps are the most fragile in the guide** — none are `skippable` and none set `doIt: false`, so they all actually click, and they match on translated text. Confirmed localized in the shipped package:

```js
label: t("grafana-prometheus.querybuilder...label-builder", "Builder")
label: t("grafana-prometheus.querybuilder...label-code", "Code")
```

Root cause was a gap in `grafana-ui`, not this guide: `RadioButtonGroup` gave every option the same static `data-testid radio-button`, with random `useId` ids and `value="on"`, so translated text was the only discriminator. grafana/grafana#129669 adds `RadioButton.option(value)` keyed on the stable enum value. `:contains()` is also non-standard CSS, so removing it is a double win.

Note on **Query inspector**: the selector `QueryTab.queryInspectorButton` has existed since 13.1.0, but was only ever consumed by the dashboard panel editor — Explore's own button never wired it up. #129669 fixes that.

### Left unchanged — waiting on 13.2.0

| Step | Current selector | Adopt once 13.2.0 ships |
| --- | --- | --- |
| Select `grafanacloud-play-prom` / `-logs` / `-traces` | `div[role='option'][aria-label='<name>']` | `data-testid data source card <name>` |

`DataSourcePicker.dataSourceCard(name)` exists, but its *parameterized* form is 13.2.0-only; before that it resolves to a shared, non-parameterized `data-source-card` for every card, so you can't pick one by name. Worth switching when available: `role="option"` and `aria-label` on that element are **conditional on `enableKeyboardNavigation`** (`VirtualizedList.tsx:105-115`), whereas the card's `data-testid` is set unconditionally.

### Cannot be fixed from grafana/grafana — where each one actually lives

Prometheus, Loki and Tempo datasource frontends have all been **externalized out of grafana/grafana** (Loki removed in grafana/grafana#129092, Tempo in grafana/grafana#127056, Prometheus now consumed as the `@grafana/prometheus` npm package). Selectors for the following need PRs against the repos below, so these steps stay as-is.

Repos resolved from each package's `repository` field and from the plugin catalogue on a live Cloud instance; file paths resolved from the shipped sourcemaps, so they are the real source paths rather than bundle names.

| Step | Upstream repo | Source file |
| --- | --- | --- |
| Prometheus **Operations** button | [grafana/grafana-prometheus-datasource](https://github.com/grafana/grafana-prometheus-datasource) (`packages/grafana-prometheus`) | `src/querybuilder/shared/OperationList.tsx` |
| Prometheus **Open metrics explorer** book icon | same | `src/querybuilder/components/MetricCombobox.tsx` |
| `div[data-testid='prometheus-dimensions-filter-item']` | same | `src/querybuilder/components/LabelFilterItem.tsx` |
| Tempo **Edit in TraceQL** button | [grafana/grafana-tempo-datasource](https://github.com/grafana/grafana-tempo-datasource) | `SearchTraceQLEditor/TraceQLSearch.tsx` |
| `div[data-testid='operations.0.wrapper']` | [grafana/plugin-ui](https://github.com/grafana/plugin-ui) (`@grafana/plugin-ui` v0.13.1) | `src/components/VisualQueryBuilder/components/OperationEditorBody.tsx` |
| `button[data-testid='install-quickpizza']` | **not public** — see below | — |
| Loki **Label browser** button, Loki **Options** row | [grafana/grafana-loki-datasource](https://github.com/grafana/grafana-loki-datasource) | Label browser already has a raw `label-browser-button`; the Options row has no test id at all |

**`install-quickpizza`** is in the `grafana-demodashboards-app` plugin (v1.18.0, the app behind `/a/grafana-demodashboards-app`). Its plugin manifest advertises **no** `links`, and `grafana/grafana-demodashboards-app` / `grafana/demodashboards-app` both 404, so there is no public repo to open a PR against — this one needs whoever owns the Cloud demo-dashboards app. Its `install-quickpizza` test id is already stable and unlocalized, so there is no urgency.

### What each fix actually involves

Both `@grafana/prometheus` and `@grafana/plugin-ui` take `@grafana/e2e-selectors` as a **peer dependency**, and the selector *values* are defined in grafana/grafana. That splits the work into three very different sizes:

| Kind of fix | PRs needed | Ships when |
| --- | --- | --- |
| Wire an **existing** selector onto an element | **1** — plugin repo only | next plugin release |
| Add a **new shared** selector | **2** — grafana/grafana defines the value, then the plugin repo consumes it | after both land, and the plugin's peer-dep floor allows the Grafana version that has it |
| Add a **raw** test id (not via the package) | **1** — plugin repo only | next plugin release, but it is not version-resolved and not discoverable from the shared package |

Per item:

| Step | Fix | Size |
| --- | --- | --- |
| **Open metrics explorer** | The selector already exists — `components.DataSource.Prometheus.queryEditor.builder.metricsExplorer` (`data-testid metrics explorer`, since 10.4.0). `MetricCombobox` just never puts it on the button; it sets only `tooltip` and `aria-label`. Add `data-testid={…metricsExplorer}` to that `Button`. | **1 PR** — smallest win here |
| **Tempo Edit in TraceQL** | No test id, no `aria-label`, no `title` — the only handle is translated button text, so this is the most fragile of the set. Needs a new selector defined in grafana/grafana, then consumed in `TraceQLSearch.tsx`. | 2 PRs |
| **Prometheus Operations** | No test id; only a translated `title="Add operation"`. Same shape as above — new selector, then wire it in `OperationList.tsx`. | 2 PRs |
| **Loki Options row** | No test id *at all* — it is a bare `<h6>` with nothing identifying in six levels of ancestry, which is why this step is markdown in the guide rather than interactive. Needs a test id adding to the collapsible. | 1–2 PRs |
| `prometheus-dimensions-filter-item` | Already a stable, unlocalized raw test id. Optional tidy-up: route it through the shared package for consistency. No behavioural gain. | 2 PRs, low value |
| Loki `label-browser-button` | Same — stable raw test id, works today. Optional tidy-up only. | 2 PRs, low value |
| `operations.0.wrapper` | Works, but the index is an array position and the params inside share a generic `autosize-input`, so a guide can only reach "the first input in operation 0". A param-level test id — `operations.${index}.param.${paramIndex}`, or better, keyed on the operation id rather than its position — would make individual params addressable. | 1 PR (plugin-ui) |
| `install-quickpizza` | Nothing to do. Already stable and unlocalized; listed only because ownership is unclear. | none |

**Priority, if someone picks this up:** *Edit in TraceQL* and *Operations* are the two that depend on translated text, so they are the only ones that actually break in a non-English Grafana. *Open metrics explorer* is worth doing simply because it is a one-line change. Everything else is cosmetic — the raw test ids are unlocalized and stable, so they carry none of the i18n fragility that motivated this pass.

<details>
<summary><b>Note on <code>operations.0.wrapper</code></b> — verified live, works today but is positional</summary>

Checked against a live Grafana Cloud 13.2.0 instance with `grafanacloud-play-logs`. `operations.0.wrapper` is the boxed **"Line contains"** card; on a fresh load `div[data-testid='operations.0.wrapper'] input` matches exactly **1** input (`placeholder="Text to find"`), so the `doughs` fill lands correctly.

Two caveats, neither currently breaking:

1. **The `0` is an array index, not an identity.** The only test ids inside are `input-wrapper`, `autosize-input`, `operations.0.add-rest-param` and icon ids — nothing marks this as *the line filter*. If Loki's builder ever stops seeding a default line filter, the step would silently type into a different operation rather than fail.
2. **The inner `input` is matched structurally.** Clicking the adjacent `+` (`operations.0.add-rest-param`) produces a **second** input with an identical `placeholder` and `autosize-input` test id, making `... input` ambiguous. The guide never clicks `+`, so this doesn't trigger in practice.

Left as-is deliberately. `:nth-match(1)` is not valid CSS (`querySelectorAll` throws), and the only standard-CSS alternative that disambiguates — `wrapper > div > div:first-child input` — hard-codes `@grafana/plugin-ui`'s DOM nesting, trading a hypothetical fragility for a real one. A proper fix needs a param-level selector upstream in `@grafana/plugin-ui`.
</details>

### Already on package selector values (no change needed)

`data-testid Select label`, `data-testid loki options`, `Panel header *`, `viz-layout-legend`, `SpanBar--wrapper`, `RadioGroup container`, `radio-button`, `Data link`, `Nav menu item`, `TimePicker Open Button`, `RefreshPicker run/interval button`, `Data source picker select container`, `Select a data source`, `Query editor row *`, `Duplicate query`, `query-history-button`, `prometheus options`, `prometheus type`, `prometheus explain switch wrapper`, `QueryEditorModeToggle`, `metric select`, `Query patterns`, and all `explore-toolbar-*`.

**Net: 5 steps now on already-shipped e2e selectors; 10 pending grafana/grafana#129669; 3 pending 13.2.0's parameterized data-source card.**

explore-getting-started/content.json
```diff
@@ -0,0 +1,785 @@
...
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "grafana:pages.Explore.toolbar.split",
+          "requirements": ["on-page:/explore", "exists-reftarget"],
+          "content": "Click **Split** to open a second, independent query pane."
+        },
+        {
+          "type": "markdown",
+          "content": "Split view is ideal for comparing two queries, two data sources, or two time ranges side by side."
```
tacole02: The "Split view is ideal…" markdown sits inside the section, so Pathfinder numbers it as a step. Same pattern later with the rate note and the Service Graph node/edge blurb. Move those outside the section (or fold them into the step tooltip/content) so they read as context, not actions.

explore-getting-started/content.json
```diff
@@ -0,0 +1,785 @@
...
+        {
+          "type": "markdown",
+          "content": "Browse history by day, star favorites, or filter by data source."
+        },
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "grafana:components.QueryTab.queryHistoryButton",
+          "requirements": ["on-page:/explore", "exists-reftarget"],
+          "content": "Click **Query history** again to close the panel."
```
tacole02: When Query history is already open, Show me / Do it on the close step feels backwards (it un-expands). Same issue on the Query inspector close. Worth making these skippable, or guiding the open state so the close step only appears when the panel is actually open.

explore-getting-started/content.json
```diff
@@ -0,0 +1,785 @@
...
+          "content": "**Label filters** let you scope a metric to exactly the series you care about."
+        },
+        {
+          "type": "interactive",
+          "action": "button",
+          "reftarget": "Operations",
+          "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
+          "doIt": false,
+          "tooltip": "This is the same function library PromQL uses — rate, sum, histogram_quantile, and dozens more — applied visually.",
+          "content": "**Operations** adds functions to your query without typing PromQL by hand."
```
tacole02, line 287: The Operations step did not highlight anything useful for me. The button-text target may not be resolving cleanly with doIt: false. Worth tightening the selector (or turning it into a clearer show-me on a concrete control).

explore-getting-started/content.json
```diff
@@ -0,0 +1,785 @@
...
+          ]
+        },
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "div[data-testid='data-testid Query editor row prometheus A']",
+          "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
+          "doIt": false,
+          "tooltip": "Each clause gets its own numbered, plain-language explanation — useful while you're still learning PromQL.",
+          "content": "Explain mode annotates every part of the query above the editor."
```
tacole02: The Explain annotations step highlighted the whole Prometheus query row and felt off. Since the previous guided step already turns Explain on, this reads better as result prose on that step (or a short markdown after it) rather than its own interactive.

explore-getting-started/content.json
```diff
@@ -0,0 +1,785 @@
...
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "grafana:components.RefreshPicker.runButtonV2",
+          "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
+          "content": "Click **Run query**."
+        },
+        {
+          "type": "markdown",
+          "content": "A counter like this only ever goes up, so it's more useful as a rate. Wrap it in `rate(...)` to see requests per second instead."
```
tacole02: The rate/rate(...) note showed up as its own numbered step. I would fold that sentence into the following formfill so the learner gets the why and the replace action together.

explore-getting-started/content.json
```diff
@@ -0,0 +1,785 @@
...
+        },
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "grafana:components.Panels.Panel.title:Raw",
+          "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-prom", "exists-reftarget"],
+          "doIt": false,
+          "skippable": true,
+          "tooltip": "Toggle to Table for a spreadsheet-style view of the exact same data.",
+          "content": "**Raw** shows the underlying data frame — every returned series and its labels."
```
tacole02, line 398: The Raw highlight covered too much for me, including the Stats tab under it. Can we tighten the target to just the Raw control/header?

explore-getting-started/content.json
```diff
@@ -0,0 +1,785 @@
...
+              "reftarget": "grafana:components.DataSourcePicker.inputV2",
+              "targetvalue": "grafanacloud-play-logs",
+              "requirements": ["exists-reftarget"]
+            }
+          ]
+        },
+        {
+          "type": "guided",
+          "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-logs"],
+          "content": "Make sure the query editor is in **Builder** mode — the next few steps describe the visual builder.",
```
tacole02, line 456: The "make sure you are in Builder mode" guided step did not work on my pass, and I could not finish the logs section from there. This one blocked progress for me.

explore-getting-started/content.json
```diff
@@ -0,0 +1,785 @@
...
+          "content": "**Span Name**, **Status**, **Duration**, and **Tags** filter further the same way."
+        },
+        {
+          "type": "interactive",
+          "action": "button",
+          "reftarget": "Edit in TraceQL",
+          "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-traces", "exists-reftarget"],
+          "doIt": false,
+          "tooltip": "Every filter you set in the form above is expressed as a single TraceQL expression here.",
+          "content": "**Edit in TraceQL** drops you into the raw query behind your form selections."
```
tacole02: Edit in TraceQL highlighting looked wonky in Block Editor. Worth a selector pass on that control.

explore-getting-started/content.json
```diff
@@ -0,0 +1,785 @@
...
+        },
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "grafana:components.Panels.Panel.title:Table - Traces",
+          "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-traces", "exists-reftarget"],
+          "doIt": false,
+          "skippable": true,
+          "tooltip": "Each row is one trace: a single end-to-end request across every service it touched.",
+          "content": "Matching traces list as a table, with trace ID, start time, service, name, and duration."
```
tacole02, line 667: The traces table highlight felt off and also pulled in the Stats tab. Same "too wide" issue as Raw in metrics.

explore-getting-started/content.json
```diff
@@ -0,0 +1,785 @@
...
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "grafana:components.RefreshPicker.runButtonV2",
+          "requirements": ["on-page:/explore", "has-datasource:grafanacloud-play-traces", "exists-reftarget"],
+          "content": "Click **Run query** to build the graph."
+        },
+        {
+          "type": "markdown",
+          "content": "Each node shows request rate and error rate for that service, and each edge is a dependency between two services — a fast way to spot where errors or latency are concentrated across a whole system."
```
tacole02: The Service Graph node/edge explanation is numbered as a step. Move it outside the section (or into the prior Run query step) so it is not treated as an action.

tacole02 (commented):
I smoke-tested this on learn.grafana.net in Block Editor (with a Playwright DOM pass first). The Explore tour and the metrics/logs/traces flows are a strong start, and the grafana: selectors mostly resolve well. I left a few notes on the diff around numbered prose, toggle/close steps, a couple of broad highlights, and the Builder mode step that blocked the logs section for me.

explore-getting-started/content.json
Jayclifford345: Fixed — and applied to every in-section markdown block, not just these three.

`AGENTS.md` rule 14 already says this ("Pathfinder often numbers in-section markdown as a step"), and the mechanism is `section-numbering.tsx` — `shouldNumberSectionChild()` returns `true` for everything except image/video/conditional, so *any* markdown inside a section gets a CSS counter number. There were **12** of them, so I did the lot rather than leave nine identical cases for the next pass:

- Split-view note → tooltip on the **Split** step
- Query-history / Query-inspector blurbs → folded into new `guided` blocks (see the thread on line 174)
- `rate()` note → into the following formfill (thread on line 364)
- Loki curly-braces note → into the **Label filters** step content
- json-parser hint, log-line-menu list, span-details list, Overview note → tooltips on their adjacent steps
- Loki **Options** row and the Service Graph node/edge blurb → moved outside the section

Every section is now steps only. Kept rule 14 intact too — the moved-out prose sits *after* each section’s one-sentence summary so the intro/summary bookends stay immediately adjacent to their section.

explore-getting-started/content.json
Jayclifford345: Fixed by restructuring rather than just flagging them skippable.

Both open/read/close trios are now a single `guided` block with two steps — which is the idiom this guide already uses for Loki `Live` tailing, so it stays internally consistent:

```json
{
  "type": "guided",
  "content": "**Query history** lists every query you have run recently — browse by day, star favorites, or filter by data source. Open it, take a look, then close it again.",
  "steps": [
    { "action": "highlight", "reftarget": "grafana:components.QueryTab.queryHistoryButton", "description": "Click Query history to open the panel" },
    { "action": "highlight", "reftarget": "grafana:components.QueryTab.queryHistoryButton", "description": "Click it again to close the panel" }
  ]
}
```

A `guided` block waits for you at each step instead of offering Show me / Do it on a standalone "close" step, so the close only comes up once you have actually opened the panel — which is the "guide the open state" option you suggested. It also removes the two markdown blocks that sat between open and close.

Separately, the inspector step was `action: button` / `"Query inspector"`. `findButtonByText()` scans `document.querySelectorAll("button")` across the **whole page including Pathfinder’s own sidebar**, so it is retargeted to `button[aria-label=Query inspector]` (verified: 1 match, 154x32). `components.QueryTab.queryInspectorButton` exists but resolves to `data-testid Query inspector button`, which Explore does not set — 0 matches live, confirming your note that grafana/grafana#129669 is still needed here.

explore-getting-started/content.json
Jayclifford345: Fixed. You were right that the button-text target was the problem, though not quite in the way it looked.

`reftarget: "Operations"` with `action: button` goes through `findButtonByText()`, which walks **every `<button>` in the document — including Pathfinder’s own sidebar** ("we want to look outside the docs plugin frame"). It takes exact text matches first, then falls back to *partial* matches. In the Block Editor your guide content is on the page, so a partial match on a sidebar element can win and you get a highlight over a chunk of the panel rather than the control.

Now scoped and switched to `highlight` (the step is `doIt: false`, so it never needed the click path):

```
div[data-testid=data-testid Query editor row prometheus A] button[title=Add operation]
```

Verified live: 1 match, 131x32, correct button. For the record the control itself is fine and fully visible — unlike Edit in TraceQL it does not overflow its container — so this was purely a resolution problem.

explore-getting-started/content.json
Jayclifford345: Fixed exactly as suggested — the standalone step is gone and its sentence is now result prose on the guided step that turns Explain on:

> Turn on **Explain** to see your query broken down in plain English. Every clause then gets its own numbered, plain-language explanation above the editor — useful while you are still learning PromQL.

That also removes the `div[data-testid=data-testid Query editor row prometheus A]` highlight, which was the whole query row and was never going to feel targeted.

explore-getting-started/content.json
Jayclifford345: Fixed — folded into the following formfill so the why and the replace action arrive together:

> **content:** Wrap the counter in `rate(...)` to see requests per second instead — replace the query with `rate(quickpizza_server_http_request_duration_seconds_count[5m])`.
> **tooltip:** A counter like this only ever goes up, so it is more useful as a rate.

explore-getting-started/content.json
Jayclifford345: Fixed, and this one had a concrete root cause worth recording.

`components.Panels.Panel.title:Raw` resolves to `data-testid Panel header Raw`, but in Explore that attribute sits on the **`<section>` wrapping the entire panel**, not on the header bar — measured live at **689x689**, of which 647px is the results table. That is why it swallowed everything below it, including the inspector’s Stats tab.

Scoped to the header container:

```
{grafana:components.Panels.Panel.title:Raw} {grafana:components.Panels.Panel.headerContainer}
```

**689x689 → 687x40**, 1 match. That box is exactly the title plus the Table/Raw toggle, which is what the step’s content and tooltip actually talk about. `Panels.Panel.headerContainer` has shipped since 9.5.0, so there is no version risk.

Also worth knowing: **Raw only exists when the query type includes Instant.** With a range-only query there is no `Panel header Raw` at all. This stack’s default new-pane query is `range + instant`, which is why you saw it. I added a `hint` for the range-only case since the step is `skippable`.

I deliberately did *not* apply this to the **Graph** and **Logs volume** panels — their step text is about the visualisation itself, so the full-panel highlight is right there.

explore-getting-started/content.json
Jayclifford345: Made non-blocking, but I should be straight with you: **I could not reproduce the resolution failure**, so I fixed the ability for it to block rather than a confirmed broken selector.

What I verified live, in both editor modes:

- `div[data-testid=data-testid Query editor row loki A]` → 1 match
- the two radio pills resolve → `Builder` (57x18) and `Code` (47x18)
- `exists-reftarget` routes through `querySelectorAllEnhanced`, which handles `:contains()`, so the requirement does not throw

So the target was reachable. What I *did* confirm is the state that makes this step load-bearing: `editorMode` lives on the query object, so when the metrics section leaves Prometheus in **Code** mode and you switch to Loki, **the Loki editor opens in Code mode** — I verified `Code` is the checked radio. This step therefore has real work to do, and if it fails to complete the rest of the section describes a builder that is not on screen. That matches "could not finish the logs section from there".

Changes:
- `skippable: true` on this block and the Prometheus equivalent, so an already-correct toggle can never wedge a section (your remedy from the line-174 thread, applied here too)
- `:contains(Builder)` → `:has(input[title=Builder])` — standard CSS instead of Pathfinder’s non-standard fallback, still 1 match
- tooltip now says why it normally needs a click, and to skip if Builder is already selected

If you can still wedge it after this, a console log of the `exists-reftarget` result at that step would pin it down.

explore-getting-started/content.json
Jayclifford345: Removed rather than reselected — "wonky" turned out to be geometry, not resolution, and there is nothing better to target.

The button resolves fine (1 exact text match). The problem is layout: measured live it sits at **x 1115-1220, inside a query-editor row that ends at x 1184**. Its immediate parent is 691px wide while the row is 647px, so the button hangs ~36px past the editor column — and the narrower the pane, the worse it gets. In the Block Editor the sidebar is taking that width, which is why it looked wrong there specifically.

There is also no handle to tighten to. The full element is:

```html
<button class="css-1nuwwf8-button" type="button" aria-disabled="false">
  <span class="css-1riaxdn">Edit in TraceQL</span>
</button>
```

No `data-testid`, no `aria-label`, no `title` — translated text is the only discriminator, which is what my own audit comment above flagged as the most fragile item in the guide (2 upstream PRs: grafana/grafana for the value, grafana/grafana-tempo-datasource to consume it).

So a selector pass cannot fix either problem today. Its one sentence now lives on the Service Name step, where it is arguably more useful because that is the form it is describing:

> **Span Name**, **Status**, **Duration**, and **Tags** filter further the same way, and **Edit in TraceQL** turns whatever you set in the form into the equivalent raw TraceQL expression.

Happy to restore it as a real step once `grafana-tempo-datasource` gives that button a test id.

explore-getting-started/content.json
Jayclifford345: Fixed, same root cause and same fix as the Raw thread — `Panel header Table - Traces` is on the `<section>` wrapping the whole panel (**689x600**), not the header bar.

```
{grafana:components.Panels.Panel.title:Table - Traces} {grafana:components.Panels.Panel.headerContainer}
```

**689x600 → 687x38**, 1 match, verified live.

explore-getting-started/content.json
Jayclifford345: Fixed — moved outside the section, and its key sentence also became the tooltip on the preceding **Run query** step so the explanation is there at the moment the graph renders:

> Each node shows request rate and error rate for one service and each edge is a dependency — a fast way to spot where errors or latency concentrate across a whole system.

Ordering keeps rule 14 intact: the section’s one-sentence summary still comes immediately after the section, and the moved prose follows it.

Jayclifford345:
## Review feedback addressed — all 10 threads

Pushed as `6c80798c`. Every flagged selector was verified with Playwright against a live **Grafana 13.2.0** Explore session on learn.grafana.net, driving Explore into each required state via the `panes=` URL rather than reloading the guide in the Block Editor.

### What changed

| Thread | Fix |
| --- | --- |
| Split-view / rate() / Service Graph prose numbered as steps | All **12** in-section markdown blocks folded into step content/tooltips or moved out |
| Query history / inspector close steps backwards | Each open/read/close trio → one `guided` block, matching the Loki `Live` idiom |
| Operations highlighted nothing useful | Scoped CSS instead of whole-document button-text matching |
| Explain annotations step felt off | Deleted; folded into the guided step that turns Explain on |
| Raw highlight too wide | `689x689` → `687x40` via `Panels.Panel.headerContainer` |
| Traces table highlight too wide | `689x600` → `687x38`, same fix |
| Builder-mode step blocked progress | `skippable: true` + standard-CSS pill selector |
| Edit in TraceQL wonky | Removed — no test id/aria-label/title, and it overflows its row |

### Two findings worth flagging

**1. Button-text targets scan Pathfinder’s own sidebar.** `findButtonByText()` walks every `<button>` in the document by design, and falls back to *partial* text matches. With the guide’s own content on the page, a sidebar element can win — which is the most likely explanation for the Operations highlight landing somewhere useless. Both remaining text targets (`Operations`, `Query inspector`) are now scoped CSS. Worth keeping in mind for other guides.

**2. Panel-title selectors are whole-panel, not header.** `components.Panels.Panel.title:X` resolves to the `<section>` wrapping an entire Explore results panel. For any step whose text is about the *header or its controls*, pair it with `Panels.Panel.headerContainer` (shipped since 9.5.0). Left `Graph` and `Logs volume` on the full panel deliberately — their step text is about the visualisation.

Also: **Raw only exists when the query type includes Instant.** A range-only query renders no `Panel header Raw` at all; this stack’s default new-pane query is `range + instant`, which is why it appeared. Added a `hint` for the range-only case.

### Correction to my audit comment above

My earlier audit claimed `RadioGroup.container`, `Switch.container` and `VizLegend.legend` were "already on package selector values". I briefly believed the opposite — that they did not exist and needed raw test ids. **Both readings were wrong in different ways, and the original audit was right:** all three were added upstream at **13.1.0**, and pathfinder pins `@grafana/e2e-selectors` **13.1.1**, so they resolve correctly. My local `node_modules` was stale at `13.0.1`, which is what produced the false negative. The guide keeps all three `grafana:` tokens; verified live, 1 match each.

Net: **36 `grafana:` tokens, 0 unresolvable.** `validate --strict` and `validate --package` both pass.

### Still upstream-blocked

Unchanged from my audit — these stay on raw selectors:

- `Query inspector` — `QueryTab.queryInspectorButton` resolves to `data-testid Query inspector button`, which Explore does not set (0 matches live). Fixed by grafana/grafana#129669, which is merged but 13.2.0-gated; `@grafana/e2e-selectors` has only 13.2.0 *prereleases* published so far.
- `RadioButton.option(value)` — same story; once released it replaces the four `:has(input[title=...])` leaves and removes the last dependency on translated labels.
- **New gap:** `QueryEditorRow.container` is `(refId) => "Query editor row ${refId}"` with no `data-testid `-prefixed variant, so it cannot scope a query row — the DOM uses `data-testid Query editor row loki A`. That is why the row scopes here are still raw, and it is a clean one-line grafana/grafana fix if we want it.
- The datasource-plugin selectors (Operations, metrics explorer, label browser, `operations.0.wrapper`, Edit in TraceQL) need PRs against the externalised plugin repos, not grafana/grafana.

explore-getting-started/content.json
```diff
@@ -0,0 +1,746 @@
...
+          "reftarget": "grafana:components.QueryEditorRow.actionButton:Duplicate query",
+          "requirements": ["on-page:/explore", "exists-reftarget"],
+          "doIt": false,
+          "tooltip": "Its neighbors let you view data source help, save the query, hide its results, remove it, or drag it to reorder.",
+          "content": "Icons on the query row include a duplicate-query shortcut, shown here."
+        },
+        {
+          "type": "guided",
+          "requirements": ["on-page:/explore"],
+          "content": "**Query history** lists every query you've run recently — browse by day, star favorites, or filter by data source. Open it, take a look, then close it again.",
```
tacole02, line 158: Re-tested after your guided rewrite. If Query history is already open when I hit this step, the close half still feels off, and the second guided step highlights empty space rather than the button. Worth another pass on open-state handling (or a different close target).

explore-getting-started/content.json
```diff
@@ -0,0 +1,746 @@
...
+              "reftarget": "grafana:components.QueryTab.queryHistoryButton",
+              "requirements": ["exists-reftarget"],
+              "description": "Click it again to close the panel"
+            }
+          ]
+        },
+        {
+          "type": "guided",
+          "requirements": ["on-page:/explore"],
+          "content": "**Query inspector** exposes the raw request and response. Its Stats, Query, JSON, and Data tabs show timing, the exact request sent, and the payload returned — invaluable when a query doesn't behave as expected. Open it, look around, then close it.",
```
tacole02, line 178: On re-test, this guided block highlighted parts of the "file history" dropdown instead of Query inspector. The rest of the tour section was fine. Looks like the aria-label='Query inspector' target is still landing on the wrong control when Pathfinder is open.

tacole02 (commented):
Re-smoke tested on learn.grafana.net after the feedback commit. Metrics, logs, and traces looked good, and the earlier fixes held up. Left two follow-ups on the Query history and Query inspector guided blocks in the tour section.

explore-getting-started/content.json
Jayclifford345: Fixed in `6dd0d82b` — and you were right to push back, because the guided rewrite couldn't have worked. The problem wasn't the selector.

**Query history opens on top of the button that toggles it.** Measured live on Grafana 13.2.0 at 1440x810 with the Pathfinder sidebar open:

| | box |
| --- | --- |
| Query history drawer | `1130x399 @ 320,411` |
| Query history **button** | `137x32 @ 513,427` |

The button sits 16px inside the drawer's top edge. So the close step resolved **1 correct element** the whole time — it was just physically underneath the panel it had opened. `elementFromPoint` at the button's centre returns the drawer's `Tab Settings`, not the button. That's your "highlights empty space rather than the button".

It's viewport-dependent, which is why it looked flaky: at 1920x992 the button is at y=306 and the drawer starts at y=595, so nothing overlaps and the step works. On a laptop it doesn't.

No selector on that button can fix this, so the close step now targets **the drawer's own close icon**:

```
{grafana:pages.Explore.QueryHistory.container} button:has(svg[data-testid='icon-times'])
```

That control only exists while the drawer is open — verified 1 match open, **0 matches closed** — so `exists-reftarget` gates the step on the open state for free. That's the "guide the open state" option you suggested, enforced by the DOM rather than by step ordering.

On your "already open when I hit this step" case: the open step is now `skippable: true`, so it can't wedge. And since the close step now actually completes, the block leaves Explore as it found it, so the stale-open state stops arising in normal flow.

Deliberately **not** matched on `aria-label`: `Close query history` is `t()`-wrapped (`explore.rich-history.close-tooltip`), and `Tab.title` is keyed on the translated label too, so tab test ids aren't locale-stable either. The icon's `data-testid` is the only unlocalised handle. I've opened grafana/grafana#130053 to give `TabbedContainer`'s close button a proper test id so this can stop reaching through the icon.

explore-getting-started/content.json
Jayclifford345: Fixed in `6dd0d82b`. Good news: this is the **same bug** as the Query history thread, cascading — so your two reports were one defect.

The `aria-label='Query inspector'` target was fine. It resolved to 1 correct element every time. What happened is:

1. The Query history close step couldn't be completed (button underneath its own drawer — see the other thread)
2. So the drawer **stayed open**
3. The inspector block then ran, and its target — Query inspector at `154x32 @ 658,427` — sits **inside that still-open drawer**, which spans `y=411-810`

And the drawer's left rail at that spot reads **"Filter history"**, with a "Filter queries for data source(s)" dropdown measured at `221x22 @ 483,475`. That's your "file history dropdown" — the highlight was drawn on the right coordinates, over the wrong panel.

Confirmed with `elementFromPoint`: with the history drawer open, the Query inspector button reports `occluded=true`, covered by `[data-testid QueryHistory]`.

So fixing the history close step fixes this one. The inspector block gets the same treatment for its own close step:

```
{grafana:pages.Explore.General.container} [role='tablist'] button:has(svg[data-testid='icon-times'])
```

The two scopes are naturally disjoint — RichHistory renders outside the Explore pane — so even with both drawers open each selector still matches exactly 1. Full state matrix verified live:

| state | history close | inspector close |
| --- | --- | --- |
| both closed | 0 | 0 |
| history open only | **1** | 0 |
| inspector open only | 0 | **1** |

Also worth recording: **`Query inspector` is no longer upstream-blocked.** My audit comment said `QueryTab.queryInspectorButton` resolved to a test id Explore never set. grafana/grafana#129669 has since merged and shipped — `data-testid Query inspector button` is live on 13.2.0 — so the open step moves off the localised `aria-label` onto the token.

tacole02 (approved):
LGTM! Thanks for the updates @Jayclifford345 !

PR #491, opened 2026-07-30
Document symbolic selectors and add convert-guide-selectors skill

Jayclifford345:
Separate from grafana/interactive-tutorials#489, which converts one guide. This adds the reusable
mechanism.

## The problem

Pathfinder can resolve a `reftarget` through `@grafana/e2e-selectors` instead of the author writing
out the resulting attribute value. Upstream documents this — `grafana-pathfinder-app`'s
[`docs/developer/interactive-examples/selectors-reference.md`](https://github.com/grafana/grafana-pathfinder-app/blob/main/docs/developer/interactive-examples/selectors-reference.md)
ranks `grafana:` paths **first** in its selector priority list. **None of that reached this repo.**
The syntax appears in exactly one guide (`dpm-explore-up-metric`) and in **none** of the seven
skills.

Worse, the reference docs here pointed the other way:

- [`docs/selectors-and-testids.md`](docs/selectors-and-testids.md) ranked a literal `data-testid` as
  **"most stable"** — priority #1.
- [`selector-patterns.md`](.cursor/commands/build-interactive-lj/reference/selector-patterns.md)
  asserted *"Pathfinder uses standard CSS selectors, NOT Playwright-style selectors"* and listed
  `:nth-match()` as an anti-pattern, while
  [`dashboard-selector-strategies.md`](.cursor/skills/autogen-guide-dashboard/dashboard-selector-strategies.md)
  recommended `:nth-match()` as "Strategy 1".

So both humans and agents hardcode resolved values, which pins a guide to one Grafana version.

**This is not hypothetical.** I hardcoded them while converting the Explore guide. A fresh agent
given the same task independently reached the same conclusion and cited these very docs as proof
the syntax doesn't exist:

> "there is no import/symbol syntax (confirmed in the guide repo's `docs/interactive-actions.md` and
> `docs/selectors-and-testids.md`)"

## Why it matters concretely

`components.QueryEditorRows.rows` resolves to:

| Grafana | Value | A literal `[data-testid='data-testid Query editor row']` |
| --- | --- | --- |
| 13.1.0+ | `data-testid Query editor row` | matches |
| < 13.1.0 | `Query editor row` (an `aria-label`) | **silently fails** |

The resolver emits `:is([data-testid="v"], [aria-label="v"])`, so the symbolic form matches both.

There's a second, sharper case. 105 of the 635 non-URL values in the package carry no
`data-testid ` prefix at all (`uplot-main-div`, `Explore Graph`, the TestData and time-range
values) — Grafana renders those as an `aria-label`. A hand-written `[data-testid='Explore Graph']`
matches **nothing on any version**. Only the reference matches.

## What's here

**Docs corrections** — rank symbolic references above literal test ids and say why; reconcile the
`:nth-match` contradiction; correct the "standard CSS only" claim that was being cited as evidence
against the feature; enumerate the engine's extensions in full; document the nav-menu exception;
apply the translation caveat consistently (it hits `action: "button"` and `:text()` exactly as hard
as `:contains()`).

**`convert-guide-selectors` skill** with the mechanical parts scripted, so runs are reproducible
rather than judgement calls:

| Script | Job |
| --- | --- |
| `build-selector-map.ts` | value→path map + probes for parameterized selectors + ambiguity list |
| `find-ambiguous.py` | values mapping to several paths, which must be pinned |
| `find-unmerged-paths.py` | selectors that exist only on your branch |
| `convert-reftargets.py` | rewrites reftargets preserving formatting; refuses unmerged selectors |
| `validate-paths.ts` | resolves every path at each supported version, flags divergence and below-floor paths |
| `tests/test-convert-reftargets.py` | 28 hermetic cases; no `GRAFANA_REPO` or `tsx` needed |

## Traps the skill encodes

Each of these silently breaks a guide and none is obvious:

1. **Nav menu items must stay literal.** Pathfinder regex-matches that exact shape to auto-expand a
   collapsed menu section; the resolved form defeats the fix. The regex appears twice
   (`dom-utils.ts`, `guided-handler.ts`), and a reference can't express the `[href]` discriminator
   anyway.
2. **Ambiguous values.** Ten values in the package map to more than one path at 13.2.0 — e.g.
   `data-testid prometheus type` is both the config page's dropdown and the query editor's
   Range/Instant/Both group — and a naive value→path lookup picks the config page for
   `dpm-explore-up-metric`'s Instant toggle, which is the wrong element. Upstream handles
   this by *giving up* (`exact.delete(value)`); the skill
   diverges deliberately, because a human with `--pin` can resolve what the picker can't.
3. **Merged ≠ released,** and `pending=0` ≠ nothing pending. Documented explicitly rather than left
   for the reader to infer.
4. **A defined selector may be unwired.** `QueryTab.queryInspectorButton` existed from 13.1.0 but
   was never wired into Explore.
5. **One bad token voids the whole reftarget.** `resolveEmbeddedGrafanaTokens` is all-or-nothing: a
   single unresolvable `{grafana:…}` returns the reftarget untouched, braces included, disabling the
   tokens that *were* correct. So validate every path, not a sample.

## Provenance

Everything is checked against `grafana-pathfinder-app` source, which
[CLAUDE.md](CLAUDE.md#pathfinder-source) already names as the authority:

| Claim | Source |
| --- | --- |
| Resolver contract, `:is(…)` output, version-awareness | `src/lib/dom/grafana-selector-core.ts` |
| Prefix dispatch, `panel:`, embedded-token all-or-nothing | `src/lib/dom/selector-resolver-core.ts` |
| Nav-menu auto-fix | `src/lib/dom/dom-utils.ts`, `action-handlers/guided-handler.ts` |
| `{grafana:…}` embedded form | `src/lib/dom/selector-generator.test.ts` (upstream unit tests) |

One remaining gap, documented rather than hidden: `grafana:<path>:<arg>` passes a single argument,
so 2-arg selectors like `components.Plugins.queryEditorRow` aren't expressible. `build-selector-map.ts`
now reports those under `undefinedProbes` and excludes them, mirroring the plugin's own
`buildTemplate` rejection, so they can't be adopted by accident.

## Verification

Built TDD-style: ran the task with **no** skill to capture baseline failures, wrote the skill against
those, then re-ran with it. The second pass produced 43 correct conversions, adopted **0** unmerged
selectors, and found 5 real defects in my first draft — including that the release gate was
structurally blind to its main case, which is why `find-unmerged-paths.py` exists.

Also checked: `find-ambiguous.py` correctly reports ambiguous values on a real guide; the gate blocks
unreleased selectors and converts the released one in a synthetic test; `validate-paths.ts`
automatically surfaces the `QueryEditorRows.rows` divergence above. Every `grafana:` path referenced
in these docs resolves at both 12.4.0 and 13.2.0.

### Review round (54e6655)

A review pass against the actual upstream source found three things worth calling out, all fixed:

- **The provenance was wrong.** This PR originally claimed the syntax was undocumented and derived
  the resolver contract by reading a minified bundle on a live Cloud instance. The upstream author
  docs and the un-minified resolver are both public; the bundle archaeology (and the
  "confirm with the Pathfinder team before merging" blocker) is gone, replaced by the citations above.
- **The converter skipped ~1 in 6 package values.** It hardcoded the `data-testid ` prefix into the
  value match, then reported those reftargets as "no data-testid attribute" — and *not* as unmapped,
  so a clean run could leave version snapshots in place while reporting nothing to do. It now matches
  any exact value, and `[aria-label='…']` reftargets too. The 43-conversion figure above predates
  this widening; a re-run finds more.
- **The PENDING report lied.** For a compound reftarget with one released part and one gated part,
  the file *was* rewritten while the report said "left unchanged" — and it double-reported the same
  reftarget as "no mappable attributes". That list is what the skill tells you to paste into the
  guide PR, so it now distinguishes "left entirely unchanged" from "WAS rewritten".

Plus: the skill was in no routing table (repeating the discovery failure that motivated it);
`selector-patterns.md` still ranked `data-testid` #1 in its headline table while its pro tip said
otherwise; `docs/selectors-and-testids.md` still called `data-testid` "most stable" under Performance
and its copy-paste tables were all literals — two of which were already **wrong** (the query mode
toggle was missing the prefix the package value carries, and `toggle-viz-picker` has two spellings
across the version range).

### Second review round (83eeac5)

Two files the docs corrections had missed, both of which contradicted the rest of the PR:

- **`.cursor/selector-library.mdc`** still ranked `data-testid` "most stable" and still carried the
  exact stale `Tab Related metrics` / `Tab Related logs` literals replaced elsewhere, with no mention
  of the syntax at all. Corrected, plus the `select-action_` literal now matches by prefix, "Button by
  Text (**Preferred**)" is no longer described as preferred (it is a text match with the same i18n
  failure as `:contains()`), and `:nth-match` is demoted with the lazy-render reason.
- **`dashboard-selector-strategies.md`** hand-rolled `panel:` scoping via a panel-header prefix and
  recommended `:nth-match(N)` for "the Nth chart globally" with no lazy-render caution. The uPlot rows
  now scope instead of index. The catalog, grading table, examples and derivation algorithm went with
  them — leaving those literal would have recreated the table-contradicts-guidance problem the first
  round flagged in `selector-patterns.md`.

**Two resolver behaviours found while writing those, neither previously documented, and both of which
my first draft of these files got wrong:**

1. **`panel:` and `{grafana:}` are mutually exclusive.** `resolveSelectorForVersion` tests
   `{grafana:` first and returns early, so `panel:` is never expanded and its prefix survives as
   literal text in front of a resolved `:is(…)`:

   ```
   in:  panel:CPU > {grafana:components.UPlotChart.container}
   out: panel:CPU > :is([data-testid='uplot-main-div'], [aria-label='uplot-main-div'])
   ```

   Nothing throws — the token resolution *succeeded* — so there is no `onError` either. Now trap 6.
   The converter's `panel:` guard already prevented emitting this, but reported it as "already
   symbolic" and had no test; both fixed.

2. **An empty argument is inexpressible.** `splitGrafanaPathParam` only splits when something follows
   the colon, so `grafana:components.Panels.Panel.title:` reads the trailing colon as part of the path
   and throws. Untitled dashboard panels (`data-testid Panel header ` with a trailing space) must stay
   literal — documented rather than silently mis-converted.

**`validate-paths.ts` no longer reports a false clean.** `resolveSelector` falls back to the *newest*
version key when none is `<=` the requested version, so a path introduced above your floor resolved
cleanly there to a value that DOM never carried — and since such a path usually has a single version
key, it also read as maximally stable in the divergence section. `components.ResourcePicker.card` is
defined only at `13.2.0` and reported a clean `OK` against Grafana 9.0.0. The script now recovers each
path's lowest version key from the versioned tree and reports these inline as `WARN` and under a
`BELOW FLOOR` section, with the count on the summary line. Warnings, not failures — a newer-only
selector is legitimate when the guide targets newer instances. `SKILL.md` step 3 and trap 7 said a
green run could not rule this out; it can now, so both are updated.

All 15 `grafana:` paths these docs reference report `BELOW-FLOOR WARNINGS: 0` across 12.4.0 and
13.2.0, the `QueryEditorRows.rows` divergence case still reports, a bad path still exits 1, and the
converter suite is 31 green (28 + 3 new `panel:` cases).


moxious (changes requested):
Fix these two and it's mergeable.

1. .cursor/selector-library.mdc was missed and now contradicts the corrected docs. This file (cross-referenced from dashboard-selector-strategies.md as the "quick-reference catalog of known selectors") still says data-testid attributes are "most stable" (line 12) and still lists the exact stale literals this PR just replaced elsewhere — Related metrics tab: button[data-testid="data-testid Tab Related metrics"] and the logs-tab equivalent (lines 68–69) — with no grafana: reference and no mention of the new syntax at all. Given the PR's own stated thesis is "these wrong claims propagated because nothing linked back to the source of truth," this file is a direct repeat of that exact failure mode.
2. dashboard-selector-strategies.md's uPlot chart guidance wasn't updated (line 35–36 on the PR branch): it still recommends :nth-match(N) for "the Nth chart globally" without the new caution this PR added elsewhere about lazy-rendering index instability, and doesn't mention the new panel:<title> > <rest> scoping syntax — which is precisely the mechanism this table hand-rolls via section[data-testid='...Panel header {title}'] div[data-testid='uplot-main-div'].

Neither is a functional break (nothing in the shipped skill/scripts depends on these files), but both undercut the PR's core argument, and a future author/agent following either file will reintroduce the exact anti-pattern this PR set out to eliminate.

PR #494, opened 2026-08-03
Docs: adds terms to readme

chri2547:
The purpose of this PR is to clarify Grafana Learn and learning path terminology.


tacole02:
This looks good, but should we expand the diagram to include Interactive Guides as embedded in LJs, since this is something we are beginning to support?

chri2547:
> This looks good, but should we expand the diagram to include Interactive Guides as embedded in LJs, since this is something we are beginning to support?

absolutely! working on that rn. Thank you!

moxious (changes requested):
active slack discussion still ongoing, and diagram has been revised a few times.  No real disagreements on the terminology, but I'd like to pair this PR with a pathfinder PR as well that actually puts into place the action on terminology choices.

PR #495, opened 2026-08-04
Add Get started with Grafana Assistant learning path

bonnywelsford-source:
## Summary
- Adds the `get-started-grafana-assistant-lj` Pathfinder learning path for Grafana Cloud: open/enable the Assistant plugin, send a first prompt, use `@` mentions, and find Learn mode.
- Registers CODEOWNERS for the new path package.

## Reviewer notes
- Assistant chat prompts are entered by the learner (`highlight` + `doIt: false`). Pathfinder `formfill` does not work on the TipTap chat input.
- Switching to Learn (and back) is one step: open the mode menu, then pick the mode. A separate step for the menu item fails because the item is not in the DOM until the menu is open.
- `value-of-assistant` stays in the package for the website Learning Path, but is omitted from Pathfinder `milestones` (framing). First hands-on is `open-and-enable-plugin` with `depends: []`.
- Enable checkbox / **Save** steps are skippable when Assistant is already enabled (for example on learn.grafana.net).

## Test plan
- [ ] Pathfinder CLI: `validate --packages get-started-grafana-assistant-lj`
- [ ] Block Editor smoke on learn.grafana.net: `open-and-enable-plugin`, `first-prompt`, `use-mentions`, `try-learn-mode`
- [ ] Confirm website framing step `value-of-assistant` is not required to start the in-app path
- [ ] Confirm enable steps skip cleanly when Assistant is already on

### DIFF
--- a/get-started-grafana-assistant-lj/end-journey/website.yaml
+++ b/get-started-grafana-assistant-lj/end-journey/website.yaml
@@ -1,0 +1,23 @@
+menuTitle: Destination reached!
+weight: 700
+step: 7
+layout: single-journey
+cta:
+  type: conclusion
+keywords:
+  - Grafana Assistant
+  - completion
+  - next steps
+side_journeys:
+  title: More to explore (optional)
+  heading: 'Continue practicing Assistant workflows:'
+  items:
+    - title: Work with Grafana Assistant
+      link: /docs/learning-paths/work-with-grafana-assistant/
+    - title: Navigate Grafana resources
+      link: /docs/grafana-cloud/machine-learning/assistant/guides/navigation/
+    - title: Query data
+      link: /docs/grafana-cloud/machine-learning/assistant/guides/querying/
+    - title: Manage dashboards
+      link: /docs/grafana-cloud/machine-learning/assistant/guides/dashboarding/
+description: Wrap up the Get started with Grafana Assistant path and explore next workflows.
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -106,6 +106,7 @@
 /visualization-metrics-lj/                                @knylander-grafana
 /visualization-traces-lj/                                 @knylander-grafana
 /git-sync-dashboards-lj/                                  @bonnywelsford-source
+/get-started-grafana-assistant-lj/                        @bonnywelsford-source
 /interactive-dashboards-lj/                               @bonnywelsford-source
 /welcome-to-play/                                         @Jayclifford345 @mdcruz
 /windows-integration/                                     @chri2547
--- a/get-started-grafana-assistant-lj/assets/selector-notes.md
+++ b/get-started-grafana-assistant-lj/assets/selector-notes.md
@@ -1,0 +1,19 @@
+# Selector notes - get-started-grafana-assistant-lj
+
+Live DOM checks on `learn.grafana.net` (2026-08-04).
+
+| Element | Selector | Notes |
+| --- | --- | --- |
+| Plugins search | `div[data-testid='input-wrapper'] input[placeholder='Search Grafana plugins']` | exists |
+| Assistant plugin card | `a[href^='/plugins/grafana-assistant-app']` | exists after search |
+| Enable checkbox | `input[type='checkbox']` (skippable) | missing when Assistant already enabled |
+| Save | button text `Save` (skippable) | missing when already enabled |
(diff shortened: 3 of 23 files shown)


bonnywelsford-source:
@grafsean can you review? This is one of the interactive learning paths going out with the intro to Grafana Assistant journey.

bonnywelsford-source:
### Why this follow-up

Reviewers (and authors) may notice this onboarding path is shorter than many learning paths. Milestone count is in a normal range for a first-win path; the real gap was **instructional depth**: steps taught *where to click* without teaching *how to decide*.

This commit keeps the path as short Cloud onboarding (enable → first chat → `@` → Learn discovery) and deepens three judgment moments instead of adding more click-through milestones or folding the content into a single interactive guide:

1. **First prompt** — criteria for a useful reply vs one that needs narrowing (handoff to `@`).
2. **Mentions** — when to mention, omit, or retry with `@`, plus what “better focus” looks like when comparing replies.
3. **Learn mode** — choose Learn for coaching that asks before changes vs Assistant for day-to-day work (title: **Choose Learn mode**).

Path framing / end-of-path outcomes were updated to match. Workflow practice (find resources, write/explain/refine queries, dashboards) stays in the companion [Work with Grafana Assistant](https://github.com/grafana/interactive-tutorials/pull/496) path.

bonnywelsford-source:
Follow-up style pass on the judgment-moment copy: no em dashes, and rewrites that avoid anthropomorphism (for example Learn/Assistant/reply as agents that pause, ask, suggest, ignore, or roam). Decision criteria are unchanged.

get-started-grafana-assistant-lj/end-journey/website.yaml
```diff
@@ -0,0 +1,23 @@
...
+keywords:
+  - Grafana Assistant
+  - completion
+  - next steps
+side_journeys:
+  title: More to explore (optional)
+  heading: 'Continue practicing Assistant workflows:'
+  items:
+    - title: Work with Grafana Assistant
+      link: /docs/learning-paths/work-with-grafana-assistant/
```
tacole02, line 16: The `/docs/learning-paths/work-with-grafana-assistant/` side journey currently 404s (companion path not published yet). Same URL is in `end-journey/content.json`. Drop or replace it until that path is live, or leave it only if you intend to ship both together.

tacole02 (approved):
Thanks for the PR. I smoke-tested this on learn.grafana.net shared in Block Editor.

The onboarding flow (enable when needed, first prompt, @ mentions, Learn vs Assistant) worked end to end on this stack. Enable steps skipped cleanly because Assistant was already on.

I left a note on the destination side journey that still points at an unpublished companion path.

bonnywelsford-source:
Thanks for the review @tacole02.  I need Sean to provide direction first before it is ready for you to  review. I should have kept in draft.

grafsean:
@bonnywelsford-source thank you for offering to generate interactive paths for my Assistant learning journey, and as we've discussed already and sharing here for context, we both agree that the interactive elements of these paths are light and mostly check if things are set up or operate the Assistant chat UI which is already very simple and straightforward.

It looks like the paths duplicate most if not all the milestones/steps of the journey (but split over two paths), is this the intent of paths? So what we've landed up with is: docs as canonical source, edited to a LJ, expanded to two IPs.

Furthermore, as we discussed some of the learning skills didn't like the length of milestones and required additional content. The journey is 2372 words, the two paths are expanded to 3151 words. For me, this changes my intent to create a simple learning resource get people started (set up and first prompts) with Assistant. 

What do you think?

bonnywelsford-source:
Thanks @grafsean I feel like one lesson is repeated and the dashboard create/expain is the only milestone that changes the job. I added decision checklists versus fake steps, which probably wasn't the right direction. 

I agree with your read.

I think the dual interactive path (Get started and work with) plan overshot what we wanted. I think I  turned a simple “set up and send first prompts” resource into two IPs that mostly restage the journey, with light chat UI interactivity and more words than the journey itself. That wasn’t the right expansion.

What about keeping one Get started path? Enable Assistant, send. a first prompt, use at-mentipn when you know a target and try learn mode? Would that be better? Park #496 and I should stop growing the path copy to satisfy milestone-heuristics? Would that work? I am unsure if would be valuable to keep it focused?

grafsean:
@bonnywelsford-source I think that's a good suggestion, feels simpler and more like what I had in mind for a quick way to get started!

PR #499, opened 2026-08-05
Fix broken YAML in read-a-trace-lj website.yaml

jdbaldry:
Will open a PR to add validation in CI to ensure this doesn't happen again


jdbaldry:
Actually there's a couple places to fix this

PR #377, opened 2026-06-03
Add first-rule learning path packages

JohnnyK-Grafana:
Pathfinder content for interactive Learning Paths for the introductory [Alerting Learning Journey.](https://github.com/grafana/website/pull/32777)
8/4: Silences content merged with Code path.

PR #502, opened 2026-08-06
Add first-rule and query-language alerting learning paths

JohnnyK-Grafana:
Replacement for #377 — same content, rebased onto current `main` as a single signed commit so the branch can pass the org signed-commits ruleset.

## Summary
- Adds `alerting-first-rule-lj` (Builder-mode first alert rule path)
- Adds `alerting-query-language-lj` (Code-mode path, including silence)
- Contact-point selectors match the live UI (`data-testid alert-rule contact-point-picker`)
- CODEOWNERS entries for both packages

## Why replace #377
#377 is 376 commits behind `main` and cannot be updated/rebased in place: merging or force-pushing would introduce unsigned commits already on `main`, which the signed-commits ruleset rejects on that branch. This PR is `main` + one signed commit.

Closes #377

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -28,6 +28,8 @@
 /adaptive-metrics-lj/                                     @brendamuir
 /adaptive-metrics-recommendations/                        @brendamuir
 /alerting-101/                                            @tomglenn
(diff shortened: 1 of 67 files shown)


tacole02 (approved):
Great work @JohnnyK-Grafana !

PR #507, opened 2026-08-12
Adds a dashboard tour guide for the Chicago L Trains dashboard on Play.

simonprickett:
This PR adds a dashboard guide for the [Chicago L Trains dashboard on Play](https://play.grafana.org/d/siwlw7f/chicago-l-trains).


leslie-heinzen (approved):
LGTM 👍🏻 🚋 

Like other play guides that have the query editor, this one fails the E2E runner because it uses a logged-in service account to test Play (which gets the new version). I ran the guide locally with playwright by working around auth and all steps succeed.

PR #510, opened 2026-08-12
Fix duplicate Install QuickPizza step on alerting paths

bonnywelsford-source:
## Summary
- Removes the redundant guided-step `description` from both alerting install-datasources milestones so docs no longer nest a second identical “Click Install the Quick Pizza SRE demo” under step 2.
- Affects `alerting-first-rule-lj` and `alerting-query-language-lj`; the guided block `content` still carries the full install instruction.

## Test plan
- [ ] After sync, confirm https://grafana.com/docs/learning-paths/alerting-query-language/install-datasources/ shows only two top-level steps with no nested duplicate.
- [ ] Same check for https://grafana.com/docs/learning-paths/alerting-first-rule/install-datasources/
- [ ] In Pathfinder, confirm the install highlight still targets `button[data-testid='install-quickpizza']` with `lazyRender`.

### DIFF
--- a/alerting-first-rule-lj/install-datasources/content.json
+++ b/alerting-first-rule-lj/install-datasources/content.json
@@ -40,8 +40,7 @@
             {
               "action": "highlight",
               "reftarget": "button[data-testid='install-quickpizza']",
-              "lazyRender": true,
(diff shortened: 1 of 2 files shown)


JohnnyK-Grafana (approved):
🔥

PR #557, opened 2026-08-27
Journey/cloud get started guides

chri2547:
This PR adds:
* 6 interactive guides meant to support Section 2 of the new Get Started with Grafana Cloud journey.
* 6 read-only interactive guides that support Section 3 of the new Get Started with Grafana Cloud journey.

(PR: https://github.com/grafana/website/pull/34353)

In section 2, each slide addresses one method of getting started with Grafana Cloud and then embeds one of guides via the interactive-guide shortcode with Try in Grafana Play = false.

Each slide in section 3 introduces one core Grafana feature and embeds one of these guides via the website's interactive-guide shortcode with the "Try in Grafana Play" button, so a brand-new user — no account, no data, no setup — can watch the feature working with real data on play.grafana.org.


tacole02:
**Scope check: does every guide here match the PR description?**

The description says all of these guides are read-only, no-account tours embedded via the website's Play shortcode ("brand-new user — no account, no data, no setup — can watch the feature working ... on play.grafana.org"). That holds for `get-started-see-alerting`, `get-started-see-dashboards`, `get-started-see-drilldown`, `get-started-see-explore`, and `get-started-see-visualizations` — their manifests correctly scope targeting to `"source": "play.grafana.org"` and their steps are `doIt:false` throughout.

The rest are not read-only Play tours:
- `get-started-connect-data` requires `has-role:admin` and performs a real "Save & test" data source creation.
- `get-started-first-check` requires `is-editor` and creates a real synthetic-monitoring check.
- `get-started-k6-test-builder` requires `is-editor` and runs a real, billed load test.
- `get-started-monitor-host` installs a real Alloy collector/integration.
- `get-started-send-data` requires `has-role:admin`, creates a real API token, and asks the user to edit their own `prometheus.yml`.
- `get-started-instrument-app` requires `has-role:admin` to create a token and to activate Application Observability.

All six target real Grafana Cloud URLs with `targetPlatform: cloud` in their manifests, with no `play.grafana.org` anchor.

Is this intentional — i.e., do these six back a different, logged-in part of the "get started" journey than the PR description covers — or should they be reworked to be read-only/Play-scoped like the others? Worth confirming before merge since it changes what review criteria should apply to them.

tacole02:
**CODEOWNERS not updated for new guide packages**

None of the guide packages added in this PR (`get-started-connect-data`, `get-started-first-check`, `get-started-k6-test-builder`, `get-started-monitor-host`, `get-started-see-alerting`, `get-started-see-dashboards`, `get-started-see-drilldown`, `get-started-see-explore`, `get-started-see-visualizations`, `get-started-send-data`) appear in `.github/CODEOWNERS`. Per the guide-PR review protocol, new guide directories should get an entry pointing at the author's handle.

get-started-see-explore/content.json
```diff
@@ -0,0 +1,81 @@
...
+          "content": "Notice the **data source** picker.\n\nIt chooses where the query runs. A data source is a connection to a place that already holds your data. To answer your teammate, you'd pick the one that collects the service's response times.",
+          "tooltip": "Every query runs against one of your connected data stores. This picker chooses which one.",
+          "requirements": ["on-page:/explore", "exists-reftarget"]
+        },
+        {
+          "type": "interactive",
+          "action": "navigate",
+          "reftarget": "/explore?schemaVersion=1&panes=%7B%22one%22%3A%7B%22datasource%22%3A%22grafanacloud-prom%22%2C%22queries%22%3A%5B%7B%22refId%22%3A%22A%22%2C%22expr%22%3A%22sum%28rate%28rpc_server_duration_milliseconds_sum%7Bservice_name%3D%5C%22productcatalogservice%5C%22%7D%5B5m%5D%29%29%20%2F%20sum%28rate%28rpc_server_duration_milliseconds_count%7Bservice_name%3D%5C%22productcatalogservice%5C%22%7D%5B5m%5D%29%29%22%2C%22datasource%22%3A%7B%22type%22%3A%22prometheus%22%2C%22uid%22%3A%22grafanacloud-prom%22%7D%7D%5D%2C%22range%22%3A%7B%22from%22%3A%22now-1h%22%2C%22to%22%3A%22now%22%7D%7D%7D",
```
tacole02, line 53: The encoded `navigate` URL sets the query's datasource `uid` to `grafanacloud-prom`, but the `content` text right below (line 54) tells the reader it runs against **`grafanacloud-play-prom`**.

Other guides that reference Play's Prometheus source use `grafanacloud-play-prom` (e.g. `alerting-first-rule-lj/build-query/content.json`, `understanding-the-four-golden-signals-of-observability/content.json`), so `grafanacloud-prom` here looks like the wrong uid — the query may fail to load or resolve to an unexpected data source on play.grafana.org. One of the two values needs to change to match the other.

get-started-connect-data/content.json
```diff
@@ -0,0 +1,115 @@
...
+    },
+    {
+      "type": "markdown",
+      "content": "First, add the connection. You need the **Admin** role in your stack:"
+    },
+    {
+      "type": "section",
+      "id": "add-your-data-source",
+      "title": "Add your data source",
+      "requirements": ["has-role:admin"],
```
tacole02, line 17: has-role:admin here means this literally can't run on Play — anonymous Play visitors only get Viewer access, never Admin. Checked the other Play-targeted guides in this repo and none of them gate on any role — they're all written assuming Viewer-only.

get-started-connect-data/content.json
```diff
@@ -0,0 +1,115 @@
...
+          "doIt": false,
+          "reftarget": "input[data-testid='search-input-input']",
+          "content": "Search for the system that holds your data, for example InfluxDB or CloudWatch, then select it from the results.",
+          "tooltip": "Type the name of the system you already run. The results filter as you type.",
+          "requirements": ["on-page:/connections/add-new-connection", "exists-reftarget"]
+        },
+        {
+          "type": "interactive",
+          "action": "button",
+          "reftarget": "Add new data source",
```
tacole02, line 45: Tested this against the two examples the previous step suggests (InfluxDB, CloudWatch) and this step breaks for both: InfluxDB's page shows **Install**, not "Add new data source" (it's a plugin that needs installing first), and CloudWatch's page doesn't have a matching element at all. The button text/flow isn't uniform across data source types, so a hardcoded `"Add new data source"` target won't hold up — worth checking whether this needs a different selector, a conditional step per data-source type, or just different example data sources that actually share this flow.

get-started-first-check/content.json
```diff
@@ -0,0 +1,171 @@
...
+        },
+        {
+          "type": "markdown",
+          "content": "Select one or more probe locations. Pick places near your users, plus one far away to see the difference distance makes."
+        },
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "button[data-testid='checkEditor form submit']",
+          "content": "Click **Save**. The check starts running on its schedule (every minute by default) and its dashboard opens.",
```
tacole02, line 117: Tested this step and it doesn't work: "Show me" highlights nothing, so `button[data-testid='checkEditor form submit']` may not match the current UI. "Do it" throws a validation error that `grafana.com` isn't a valid host — which traces back to the target value this guide tells you to enter two steps earlier (line 98). So even with a matching selector, submission would still fail on the example value the guide itself suggests. Same pattern as the InfluxDB/CloudWatch issue on `get-started-connect-data` — worth checking both the selector and swapping in a target value that actually validates for a Ping check.

tacole02: I couldn't complete testing this guide after getting stuck here.

get-started-k6-test-builder/content.json
```diff
@@ -0,0 +1,167 @@
...
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "a[data-testid='data-testid Nav menu item'][href='/testing-and-synthetics']",
+          "content": "Click **Testing & synthetics** in the main menu.",
+          "requirements": ["navmenu-open", "exists-reftarget"]
+        },
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "a[data-testid='data-testid Nav menu item'][href='/a/k6-app']",
```
tacole02, line 29: Tested this step: clicking through causes the left nav to visibly bounce/reflow before it settles on the right target. `navmenu-open` is set, but this item may sit in a submenu that needs its own expand step, or the highlight is firing before the mega-menu finishes its own animation/reflow. Worth checking if there's a missing wait/expand condition here.

get-started-monitor-host/content.json
```diff
@@ -0,0 +1,140 @@
...
+    },
+    {
+      "type": "markdown",
+      "content": "Time to add the prebuilt dashboards."
+    },
+    {
+      "type": "section",
+      "id": "install-the-dashboards",
+      "title": "Install the dashboards",
+      "requirements": ["section-completed:set-up-the-collector", "on-page:/connections/add-new-connection"],
```
tacole02, line 82: This section appears to require Admin/Editor role (installing dashboards and alert rules needs elevated permissions in Grafana Cloud) but has no role requirement gating it. Same Play-compatibility concern as the other guides flagged above — this wouldn't be supported on Play either.

get-started-monitor-host/content.json
```diff
@@ -0,0 +1,140 @@
...
+          "action": "highlight",
+          "reftarget": "button[data-testid='install-button']",
+          "content": "Click **Install**. This adds the integration's dashboards and alert rules to your stack.",
+          "requirements": ["exists-reftarget"]
+        },
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "a[href^='/dashboards/f/integration']",
+          "content": "Click **View Dashboards** and open one. Your machine's metrics are already on it.",
```
tacole02, line 95: Tested this step: there's weird scrolling/jitter before it settles on the target. Possibly needs `lazyRender`/`guided` if the dashboard-folder link list is long enough to scroll, similar to the k6-test-builder nav issue.

get-started-see-drilldown/content.json
```diff
@@ -0,0 +1,112 @@
...
+          "type": "interactive",
+          "action": "navigate",
+          "reftarget": "/a/grafana-lokiexplore-app/explore",
+          "verify": "on-page:/a/grafana-lokiexplore-app",
+          "content": "Open the **Logs** section of Drilldown.\n\nMetrics are numbers, and numbers tell you when something changed. Logs are the lines a service writes while it runs, and they usually tell you why. A selected metric offers a **Related logs** tab that shows matching log lines in place, but this investigation needs the full picture: every service that writes logs, each with a chart of how much it's logging."
+        },
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "a[data-testid='data-testid button-select-service'][href*='/service/checkoutservice/']",
```
tacole02, line 91: Tested this step and the selector doesn't match the current UI — `a[data-testid='button-select-service'][href*='/service/checkoutservice/']` didn't resolve, had to get AI help to find a working one. Because the click never actually navigates, the next step's `on-page:/a/grafana-lokiexplore-app/explore/service/checkoutservice` requirement (line 103) can't be satisfied either, so it throws "Navigate to the '/a/grafana-lokiexplore-app/explore/service/checkoutservice' page first." Worth checking this against the actual Logs Drilldown service list; the testid or href pattern may have changed.

tacole02: I got stuck at this step and couldn't test the rest of this guide.

get-started-send-data/content.json
```diff
@@ -0,0 +1,189 @@
...
+    },
+    {
+      "type": "markdown",
+      "content": "Now create a token and copy the configuration. You need the **Admin** role in your stack."
+    },
+    {
+      "type": "section",
+      "id": "create-a-token-and-copy-the-config",
+      "title": "Create a token and copy the config",
+      "requirements": ["section-completed:open-the-setup-page", "has-role:admin", "on-page:/connections/add-new-connection/hmInstancePromId"],
```
tacole02, line 61: This section probably isn't runnable on Play — `has-role:admin` would block it, same Play-compatibility issue flagged on the other guides in this PR.

get-started-demo-data/manifest.json
```diff
@@ -0,0 +1,29 @@
...
+  "type": "guide",
+  "description": "Install the QuickPizza demo from the Demo Data Dashboards app and query its metrics in Explore, to see what an instrumented app's telemetry looks like in Grafana Cloud before instrumenting your own.",
+  "category": "onboarding",
+  "author": {
+    "team": "Grafana Documentation"
+  },
+  "startingLocation": "/a/grafana-demodashboards-app",
+  "targeting": {
+    "match": {
+      "and": [
```
tacole02: Ran this through `scripts/lint-targeting.py` and it flags the manifest: the root `and` mixes URL and non-URL criteria, so it scores above zero on every Cloud page instead of just `/a/grafana-demodashboards-app`. Every other manifest in this PR wraps its `and` in an `or` — looks like this one just missed it.

get-started-demo-data/content.json
```diff
@@ -0,0 +1,126 @@
...
+          ]
+        },
+        {
+          "type": "guided",
+          "content": "Click **Install the Quick Pizza SRE demo**.\n\nThis deploys the demo data sources and dashboards, then refreshes the page.",
+          "completeEarly": true,
+          "steps": [
+            {
+              "action": "highlight",
+              "reftarget": "button[data-testid='install-quickpizza']",
```
tacole02, line 37: This step's missing a `requirements` array. There's a near-identical step already in the repo for this exact button (`alerting-first-rule-lj/install-datasources/content.json`, same `install-quickpizza` testid) that has `"requirements": ["exists-reftarget"]` — probably just got dropped when this was adapted from there.

chri2547:
> **Scope check: does every guide here match the PR description?**
> 
> The description says all of these guides are read-only, no-account tours embedded via the website's Play shortcode ("brand-new user — no account, no data, no setup — can watch the feature working ... on play.grafana.org"). That holds for `get-started-see-alerting`, `get-started-see-dashboards`, `get-started-see-drilldown`, `get-started-see-explore`, and `get-started-see-visualizations` — their manifests correctly scope targeting to `"source": "play.grafana.org"` and their steps are `doIt:false` throughout.
> 
> The rest are not read-only Play tours:
> 
> * `get-started-connect-data` requires `has-role:admin` and performs a real "Save & test" data source creation.
> * `get-started-first-check` requires `is-editor` and creates a real synthetic-monitoring check.
> * `get-started-k6-test-builder` requires `is-editor` and runs a real, billed load test.
> * `get-started-monitor-host` installs a real Alloy collector/integration.
> * `get-started-send-data` requires `has-role:admin`, creates a real API token, and asks the user to edit their own `prometheus.yml`.
> * `get-started-instrument-app` requires `has-role:admin` to create a token and to activate Application Observability.
> 
> All six target real Grafana Cloud URLs with `targetPlatform: cloud` in their manifests, with no `play.grafana.org` anchor.
> 
> Is this intentional — i.e., do these six back a different, logged-in part of the "get started" journey than the PR description covers — or should they be reworked to be read-only/Play-scoped like the others? Worth confirming before merge since it changes what review criteria should apply to them.

Addressed this. Thank you.

chri2547:
> **CODEOWNERS not updated for new guide packages**
> 
> None of the guide packages added in this PR (`get-started-connect-data`, `get-started-first-check`, `get-started-k6-test-builder`, `get-started-monitor-host`, `get-started-see-alerting`, `get-started-see-dashboards`, `get-started-see-drilldown`, `get-started-see-explore`, `get-started-see-visualizations`, `get-started-send-data`) appear in `.github/CODEOWNERS`. Per the guide-PR review protocol, new guide directories should get an entry pointing at the author's handle.

added, thank you.

get-started-see-explore/content.json
chri2547: AI opened Play and tested, and this worked as is.

get-started-connect-data/content.json
chri2547: It's ok for this not to run on Play

get-started-demo-data/content.json
chri2547: That step carries lazyRender because exists-reftarget can't scroll to below-the-fold targets (critical rule 21), so gating the block on exists-reftarget would reintroduce the failure lazyRender exists to avoid.

get-started-demo-data/manifest.json
chri2547: DOne.

get-started-send-data/content.json
chri2547: No need to run this on Play.

get-started-connect-data/content.json
chri2547: Good catch — this was an example problem, not a selector problem. On a Cloud stack, searching "InfluxDB" or "CloudWatch" surfaces their integration tiles, which have Install flows instead of the plain data source page. I switched the examples to MySQL and PostgreSQL, which are core data sources that ship inside Grafana itself, so their catalog pages always offer Add new data source with no install step. I also added a recovery line to the step: if the page offers Install instead, you've opened an app or integration, go back and pick the data source result. The journey slide that names the examples was updated to match.

get-started-first-check/content.json
chri2547: This one took some digging, and both symptoms turned out to have a single upstream cause. I verified against the synthetic-monitoring-app source that the Save button's checkEditor form submit testid is current (it was added specifically for Pathfinder guides), and that grafana.com is a valid ping target — it's literally the field's placeholder constant in the app code. The real bug was two steps earlier: the Ping radio selector used data-testid radio-button-option ping, which only exists in Grafana core 13.2+. On current Cloud stacks that click silently fails, the check stays on HTTP, and grafana.com then correctly fails HTTP URL validation at save — exactly the error you saw. I replaced the radio selector with one anchored on the SM app's own group testid (checkEditor form chooseAPICheckType), which doesn't depend on the Grafana core version. With the check actually on Ping, the target value and Save button both work.

get-started-monitor-host/content.json
chri2547: converted the View Dashboards step to a guided block with lazyRender: true so the engine scrolls the link into view before highlighting, per rule 21.

get-started-see-drilldown/content.json
chri2547: Interesting one: I tested this live on Play and the selector itself is actually correct — a[data-testid='data-testid button-select-service'][href*='/service/checkoutservice/'] matches the current DOM exactly. What bit you is that the service list renders 20 of 77 services per page, ordered by log volume, so whether checkoutservice is on page 1 depends on the day. I added a step that searches for checkoutservice in the service filter first (verified live that it brings the card into view), and the click step now runs against a filtered list. That also fixes the downstream on-page requirement you hit, since the click can actually navigate now.

get-started-first-check/content.json
```diff
@@ -0,0 +1,171 @@
...
```
tacole02: Retested after the Ping-selector fix, and this step is still broken: "Show me" finds nothing for `button[data-testid='checkEditor form submit']`. Since "Show me" doesn't depend on form validity, this looks like a different bug than the one already fixed — the button just isn't in the DOM at this point.

Two other guides in this repo use this identical selector for the same Save action (`sm-ping-check-tutorial`, `detect-outages-synthetic-monitoring-lj/select-probe-locations`), and both visit an **Alerting** tab before Save — one also clicks **Test** first. This guide goes straight from **Execution** to **Save**. Worth checking whether Save only renders once you've reached the last tab of the wizard, in which case this step would need to visit Alerting first too.

get-started-see-drilldown/content.json
```diff
@@ -0,0 +1,120 @@
...
+          "reftarget": "/a/grafana-lokiexplore-app/explore",
+          "verify": "on-page:/a/grafana-lokiexplore-app",
+          "content": "Open the **Logs** section of Drilldown.\n\nMetrics are numbers, and numbers tell you when something changed. Logs are the lines a service writes while it runs, and they usually tell you why. A selected metric offers a **Related logs** tab that shows matching log lines in place, but this investigation needs the full picture: every service that writes logs, each with a chart of how much it's logging."
+        },
+        {
+          "type": "interactive",
+          "action": "formfill",
+          "reftarget": "input[data-testid='data-testid search-services']",
+          "targetvalue": "checkoutservice",
+          "content": "Search for `checkoutservice` in the service list.\n\nDozens of services write logs here, spread across several pages, so search brings the one you want to the top.",
```
tacole02: Tested live on Play. This search box isn't a list filter — typing into it leaves the results table below completely unchanged (I confirmed `checkoutservice`'s card/link is already there, position 1 of 20, before and after typing). Selecting the autocomplete suggestion instead navigates directly to the service's logs page, bypassing "Show logs" entirely. So there's no sequence of (search, then click Show logs) that actually works here — either the search step is a no-op and the next step should find the card fine on its own (checkoutservice is on page 1 right now, so the original pagination theory doesn't seem to be the blocker), or the automated fill is triggering the suggestion-select behavior and jumping past this step to the logs page before the next step runs, which is why it can't find anything. Worth checking which of those two is actually happening in the guide engine — I could reproduce the first but not conclusively the second by hand.

tacole02: Correction to my last comment — tested further and the search step isn't the problem. I called `.click()` directly on the checkoutservice "Show logs" link (search box still had text typed, dropdown still open, nothing selected) and it worked fine, navigating straight to the right logs page. The one thing I did that this step doesn't: scroll it into view first. Its default position sits just past a typical viewport's fold, and per `docs/guided-interactions.md`, a plain `interactive`/`highlight` step can't scroll to reach it — same root cause already fixed twice elsewhere in this PR (`monitor-host`, `demo-data`). This step likely needs the same treatment: convert to `guided` with `lazyRender: true`.

tacole02 (dismissed):

tacole02 (approved):
Looks great! Thanks, @chri2547 !

PR #563, opened 2026-08-31
Validate snippet references against generated catalog

Nikshay1:
## Summary

Updates `validate-json.yml` to pass the generated snippets catalog into strict `content.json` validation when the catalog exists.

The workflow already builds `/tmp/snippets-index.json` while validating snippet bodies. This change reuses that artifact by conditionally adding:

`--snippets-catalog /tmp/snippets-index.json`

to each `pathfinder-cli validate --strict` invocation.

If a repository or branch has no `shared/snippets` directory, the catalog argument remains empty and validation continues with the existing schema-only behavior.

## Why

This is the CI half of grafana/grafana-pathfinder-app#1456.

Without this check, a guide can contain a schema-valid `snippet-ref` whose `snippetId` does not exist in the generated catalog and still pass CI.

The corresponding Pathfinder CLI support is implemented in grafana/grafana-pathfinder-app#1750.

## Validation

- `git diff --check`
- Pathfinder CLI implementation passes its focused snippet-reference tests
- Pathfinder `npm run check` passes
- verified the catalog argument is only supplied when `/tmp/snippets-index.json` exists

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -62,10 +62,10 @@ jobs:
           # Bumped again from 4d697f80: that pin's `build-repository` dropped the `stats` key
           # the build-stats step above had just written, so the stamping was inert. This pin is
           # grafana-pathfinder-app#1662, which forwards unknown top-level manifest keys into
-          # repository.json — the smallest move that makes the stats reach the CDN. Deliberately
-          # stops short of #1710 (CLI command binding); it rides along on #1684, which adds the
-          # `callout` block type, counted as one non-container block like `markdown`.
-          ref: acbf39b4fbad1459a807754e2d54488e3d58df3e
+          # repository.json. Bumped again to keep deploy on the same CLI as validate-json.yml;
+          # this pin includes #1684's `callout` block type, #1702's `divider` block type, and
+          # #1750's snippet catalog validation support.
+          ref: 37d755c5f67cad6268d8ee38f64bf06313acf4b9
           path: pathfinder-app
           persist-credentials: false
 
@@ -153,10 +153,10 @@ jobs:
           # Bumped again from 4d697f80: that pin's `build-repository` dropped the `stats` key
           # the build-stats step above had just written, so the stamping was inert. This pin is
           # grafana-pathfinder-app#1662, which forwards unknown top-level manifest keys into
-          # repository.json — the smallest move that makes the stats reach the CDN. Deliberately
-          # stops short of #1710 (CLI command binding); it rides along on #1684, which adds the
-          # `callout` block type, counted as one non-container block like `markdown`.
-          ref: acbf39b4fbad1459a807754e2d54488e3d58df3e
+          # repository.json. Bumped again to keep deploy on the same CLI as validate-json.yml;
+          # this pin includes #1684's `callout` block type, #1702's `divider` block type, and
+          # #1750's snippet catalog validation support.
+          ref: 37d755c5f67cad6268d8ee38f64bf06313acf4b9
           path: pathfinder-app
           persist-credentials: false
 
--- a/.github/workflows/validate-json.yml
+++ b/.github/workflows/validate-json.yml
@@ -72,8 +72,8 @@ jobs:
         # which added `build-stats`; 2262da89 predates it and has no such command.
         # Bumped again from 4d697f80 to keep this validator on the same CLI as deploy.yml, whose
         # `build-repository` needed grafana-pathfinder-app#1662 to forward the `stats` key at all.
-        # Carries #1684's new `callout` block type, so a guide using one validates here.
-        ref: acbf39b4fbad1459a807754e2d54488e3d58df3e
+        # Carries #1684's new `callout` block type and #1750's snippet catalog validation support.
+        ref: 37d755c5f67cad6268d8ee38f64bf06313acf4b9
         path: pathfinder-app
         persist-credentials: false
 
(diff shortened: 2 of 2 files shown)


moxious (changes requested):
Thanks for wiring this up — reusing the throwaway `/tmp/snippets-index.json` instead of building a second catalog is the right call, and gating the flag on the file's existence keeps repos/branches without `shared/snippets` working exactly as before.

Two things need to change before this can merge, both needed to make the check actually run end-to-end:

1. **Bump the grafana-pathfinder-app ref pin.** `--snippets-catalog` only exists as of grafana/grafana-pathfinder-app#1750, which is now merged to `main` at `37d755c5f67cad6268d8ee38f64bf06313acf4b9`. This workflow still checks out `acbf39b4...` (predates #1750), so the flag doesn't exist in the CLI build this job runs today. Please bump `ref:` at line 76 to `37d755c5f67cad6268d8ee38f64bf06313acf4b9` (or later), and update the adjacent comment explaining the pin to mention #1750 alongside the existing #1684 callout-block reference.

2. **Wire the second `validate` call too.** Only the per-file loop (`validate --strict "${SNIPPET_CATALOG_ARGS[@]}" "$file"`) got the new flag. The package-directory loop a few dozen lines down (`validate --package "$dir"`) didn't — and this repo has 667 `manifest.json` package directories, the majority of all guides here. As submitted, this change would only check snippet-refs in single-file guides and leave the rest of the repo completely unvalidated. Please add the same `"${SNIPPET_CATALOG_ARGS[@]}"` to that second invocation.

**What to expect once both of those land, and why a red build here would actually be good news:** once the pin is bumped and both call sites are wired, this check runs for real, for the first time, against every guide in this repo. If it finds genuinely broken `snippet-ref`s that have been silently wrong until now, this PR's CI will go red — and that's a _good_ outcome, not a problem with your change: it's proof the validator does exactly what it's supposed to do. Don't try to make this PR pass by loosening the check or reducing its scope. Instead:

- Push the two fixes above and let CI run for real.
- If it reports failures, leave them as-is here — we'll open separate, targeted PR(s) against the specific guides it flags to repair those snippet references.
- Once those repairs land, this PR will go green on its own (no changes needed to this PR itself), and we'll merge it clean.

So a failing build after your update is a signal to hold this PR open while we fix the guides it caught, not a sign that something's wrong with your fix.

Nikshay1:
Thanks, fixed both. Bumped the Pathfinder pin and wired the catalog into package validation too. If CI catches any broken refs, I’ll leave those for separate fixes.

moxious (changes requested):
Nice — I verified the mechanics and they hold up: `37d755c5` is the real merge commit of grafana-pathfinder-app#1750 on `main`; `--snippets-catalog` is honoured in **both** modes this workflow uses (`runFileValidation` via `validateGuide`, and `runPackageValidation` via `validatePackage` → `snippetCatalogIds`), so passing it after `--package "$dir"` is fine; `build-snippets -o` emits exactly the `Record<id, SnippetCatalogEntry>` shape `SnippetCatalogSchema` expects, and `resolveCliPath` honours the absolute `/tmp` path. The catalog is written by the same job, `"${ARR[@]}"` empty-expansion is safe under GH's `bash -e` default, and no guide in the repo has a `snippet-ref` block today, so this won't turn CI red on merge.

Two things before I approve.

**1. The pin bump silently drops validator/deploy CLI parity (blocking).**

The comment two lines above the new pin still asserts the invariant — *"Bumped again from 4d697f80 to keep this validator on the same CLI as deploy.yml"* — but `deploy.yml` still pins `acbf39b4` at both line 68 and line 159. `37d755c5` is 44 commits ahead, and that range includes #1702, which adds the `divider` block type. I diffed the block-type literals in `json-guide.schema.ts` at both refs: `divider` exists only at `37d755c5`.

Concrete failure: someone merges a guide with a `divider` block. PR validation passes here on the new CLI. Then deploy's `build-stats` step parses `content.json` with the old `ContentJsonSchema`, rejects the unknown block type, and — because `build-stats` aborts the whole tree on any error rather than skipping the offending package — no manifests get stamped and the deploy exits 1.

The same divergence undercuts the `build-stats` / `build-repository` steps in this workflow (lines 254-273), whose comment justifies them explicitly as a fail-fast rehearsal of deploy. They now rehearse a different CLI than deploy runs.

Either bump `deploy.yml`'s two pins to `37d755c5` in this PR, or update the line-73 comment to state that the parity invariant is being dropped deliberately, and why that's safe.

**2. The catalog guard makes the new check self-disabling and silent (minor).**

```bash
if [ -f /tmp/snippets-index.json ]; then
```

If `shared/snippets` is ever renamed or removed, or the "Validate snippet bodies" step's `-o` path changes without these two blocks being updated in lockstep, `SNIPPET_CATALOG_ARGS` stays empty, `validate` quietly falls back to schema-only behaviour, and CI reports a clean green run with no signal that the snippet-reference check didn't run — shipping exactly the dangling `snippet-ref` this PR exists to catch.

Cheap fix: `echo` which mode was selected, or emit a `::warning::`, in an `else` branch, so the log distinguishes "checked" from "skipped". Same block at line 231 has the same issue.

Nikshay1:
Thanks for the careful review, @moxious. Addressed both follow-ups in `3d8b93ca`: all deploy/validation CLI pins now use `37d755c5`, and both catalog guards clearly log checked vs. skipped. I also ran the generated catalog through all 667 guides in both strict file and package modes—667/667 passed in each. Appreciate you catching the parity and observability gaps; ready for another look when convenient.

moxious (approved):
Thanks for the contribution!

PR #565, opened 2026-09-03
Add PostgreSQL data source learning path

lwandz13:
## What

Adds a new interactive learning path package **`postgresql-data-source-lj`** for connecting the PostgreSQL data source in Grafana Cloud.

Six milestones (modeled on `mysql-data-source-lj`):

1. **prepare-configuration** — create a read-only `grafanareader` user, `GRANT SELECT` (markdown)
2. **add-data-source** — Connections → Add new data source → search **PostgreSQL** → name it
3. **configure-datasource** — Host URL (`:5432`), Database name, Username, Password, TLS/SSL Mode
4. **test-connection** — select PDC → **Save & test**
5. **verify-postgres-data** — Explore → run a test SQL query
6. **end-journey** — recap + related paths

Content sourced from the canonical PostgreSQL data source docs (overview, configure, query-editor, troubleshooting). Also adds a CODEOWNERS entry.

## Selectors needing verification

These were reused from the working MySQL path or are PostgreSQL-form guesses; verifying via the PR tester / Block Editor:

- `configure-datasource`: `input[id='connection-database']` (Database name)
- `configure-datasource`: `input[id='connection-sslmode']` (TLS/SSL Mode)
- `test-connection`: `input[aria-label='Private data source connect']`

## Testing

- [x] Block Editor smoke test each milestone (Show me / Do it)

### DIFF
--- a/postgresql-data-source-lj/add-data-source/content.json
+++ b/postgresql-data-source-lj/add-data-source/content.json
@@ -51,14 +51,13 @@
         },
         {
           "type": "interactive",
-          "action": "formfill",
-          "reftarget": "input[placeholder='Name']",
-          "targetvalue": "Production PostgreSQL Database",
+          "action": "highlight",
+          "reftarget": "[data-testid='data-testid Editable title edit button']",
           "doIt": false,
           "requirements": [
             "exists-reftarget"
           ],
-          "content": "In the **Name** field, enter a descriptive name for your PostgreSQL data source.\n\nFor example, enter `Production PostgreSQL Database`.\n\n**Did you know?** The name of the data source is displayed on dashboard panels, so it's important to choose a meaningful name."
+          "content": "To give your data source a descriptive name, click the **edit title** (pencil) icon next to the name at the top of the page, then enter a name such as `Production PostgreSQL Database`.\n\n**Did you know?** The name of the data source is displayed on dashboard panels, so it's important to choose a meaningful name."
         }
       ]
     },
--- a/postgresql-data-source-lj/add-data-source/content.json
+++ b/postgresql-data-source-lj/add-data-source/content.json
@@ -1,0 +1,74 @@
+{
+  "schemaVersion": "1.1.0",
+  "id": "postgresql-data-source-add",
+  "title": "Add the PostgreSQL data source",
+  "blocks": [
+    {
+      "type": "markdown",
+      "content": "Grafana ships with the PostgreSQL data source preinstalled in Grafana Cloud, so you don't need to install a plugin. In this milestone, you add the PostgreSQL data source and give it a descriptive name.\n\nLet's walk through the steps together."
+    },
+    {
+      "type": "section",
+      "autoCollapse": false,
+      "blocks": [
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "a[href='/connections/datasources']",
+          "requirements": [
+            "navmenu-open",
+            "exists-reftarget"
+          ],
+          "content": "In the navigation menu, click **Connections > Data sources**."
+        },
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "[data-testid='data-testid data-source-add-button']",
+          "requirements": [
+            "exists-reftarget"
+          ],
+          "content": "Click **Add new data source**."
+        },
+        {
+          "type": "interactive",
+          "action": "formfill",
+          "reftarget": "input[placeholder='Filter by name or type']",
+          "targetvalue": "PostgreSQL",
+          "requirements": [
+            "exists-reftarget"
+          ],
+          "content": "In the **Search** field, enter `PostgreSQL`."
+        },
+        {
+          "type": "interactive",
+          "action": "button",
+          "reftarget": "PostgreSQL",
+          "requirements": [
+            "exists-reftarget"
+          ],
+          "content": "From the search results, click **PostgreSQL**."
+        },
+        {
+          "type": "interactive",
+          "action": "formfill",
+          "reftarget": "input[placeholder='Name']",
+          "targetvalue": "Production PostgreSQL Database",
+          "doIt": false,
+          "requirements": [
+            "exists-reftarget"
+          ],
+          "content": "In the **Name** field, enter a descriptive name for your PostgreSQL data source.\n\nFor example, enter `Production PostgreSQL Database`.\n\n**Did you know?** The name of the data source is displayed on dashboard panels, so it's important to choose a meaningful name."
+        }
+      ]
+    },
+    {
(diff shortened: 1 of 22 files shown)


postgresql-data-source-lj/verify-postgres-data/content.json
```diff
@@ -0,0 +1,71 @@
...
+        {
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "[data-testid='query-editor-rows']",
+          "doIt": false,
+          "requirements": [
+            "on-page:/explore",
+            "exists-reftarget"
+          ],
+          "content": "Switch to **Code** mode and enter a simple SQL query to test your connection.\n\nFor example, enter:\n\n```sql\nSELECT datname, numbackends FROM pg_stat_database LIMIT 5\n```\n\n**Tip:** You can use any valid SQL query that your PostgreSQL user has permission to run."
```
tacole02: When I ran this in Block Editor, Show me highlighted the whole query panel rather than **Code** mode. Explore was already in Code mode, so the rest of the step still worked. Worth changing the copy to "In the query editor, enter..." (and skip switching modes), or adding a skippable highlight on **Code** if you still want that click.

postgresql-data-source-lj/configure-datasource/content.json
```diff
@@ -0,0 +1,72 @@
...
+          "requirements": [
+            "exists-reftarget"
+          ],
+          "content": "In the **Password** field, enter the password for the PostgreSQL user you created."
+        }
+      ]
+    },
+    {
+      "type": "markdown",
+      "content": "Next, set the **TLS/SSL Mode** to match your database. For cloud-hosted databases such as Amazon RDS, Azure Database for PostgreSQL, and Google Cloud SQL, `require` is the recommended minimum. Use `verify-full` for the most secure option, or `disable` only for local development on a trusted network."
```
tacole02, line 61: I smoke-tested configure on the PostgreSQL settings page. **TLS/SSL Mode** is on the form, but this section never highlights it, and the destination recap (and path outcomes) still say the learner set TLS/SSL. Add a `doIt: false` step for that field, or drop TLS/SSL from the recap and outcomes if it stays prose-only.

postgresql-data-source-lj/add-data-source/content.json
```diff
@@ -0,0 +1,73 @@
+{
+  "schemaVersion": "1.1.0",
+  "id": "postgresql-data-source-add",
+  "title": "Add the PostgreSQL data source",
+  "blocks": [
+    {
+      "type": "markdown",
+      "content": "Grafana ships with the PostgreSQL data source preinstalled in Grafana Cloud, so you don't need to install a plugin. In this milestone, you add the PostgreSQL data source and give it a descriptive name.\n\nLet's walk through the steps together."
```
tacole02: Opening copy says Grafana Cloud ships PostgreSQL preinstalled. The postgres docs say that for OSS and Enterprise. Safer to say it is a built-in core data source and you don't need to install a plugin.

postgresql-data-source-lj/test-connection/content.json
```diff
@@ -0,0 +1,53 @@
...
+  "id": "postgresql-data-source-test",
+  "title": "Select PDC and test connection",
+  "blocks": [
+    {
+      "type": "markdown",
+      "content": "The private data source connect (PDC) feature lets you establish a secure connection between Grafana Cloud and a PostgreSQL database on a private network.\n\nGrafana Cloud can't reach databases on `localhost`, `127.0.0.1`, or private IP ranges (`10.x`, `172.16.x`, `192.168.x`) directly. If your database isn't publicly accessible, PDC creates a secure tunnel so the connection can succeed."
+    },
+    {
+      "type": "markdown",
+      "content": "**Recommendation:** We recommend that you create a private connection whenever you connect Grafana Cloud to an external data source."
```
tacole02: This recommends a private connection whenever you connect Grafana Cloud to an external data source. Docs recommend PDC when the database isn't publicly accessible. The skippable PDC step already covers the public case.

tacole02 (commented):
Thanks for the PR. I smoke-tested this on the shared learn.grafana.net stack in Block Editor.

The path walks through preparing a read-only user, adding PostgreSQL, filling connection details, testing with PDC, and running a query in Explore. Add, configure, and test ran cleanly. Verify mostly worked. Show me on the SQL step highlighted the whole query panel because Explore was already in Code mode.

I left a few notes on the diff.

postgresql-data-source-lj/verify-postgres-data/content.json
lwandz13: Good catch. Fixed in 4f72ab80: reworded to "In the query editor, enter a simple SQL query..." and moved the mode switch into a conditional tip ("if the editor is in **Builder** mode, switch to **Code** first"), so "Show me" no longer implies a Code-toggle click. This matches the copy the `mysql-data-source-lj` sibling already uses.

postgresql-data-source-lj/configure-datasource/content.json
lwandz13: Fixed in 4f72ab80. I went with the prose-only option: the **TLS/SSL Mode** field is a Grafana UI `Combobox` with no `name`, `id`, or `data-testid`, and `input[role='combobox']` isn't unique on the page (several comboboxes), so there's no stable selector to add a reliable `doIt: false` step (the earlier `input[id='connection-sslmode']` guess wouldn't resolve). I kept the informative TLS/SSL prose but softened the claims: the destination recap now says "Reviewed the TLS/SSL mode options..." and the path outcomes say "choose a TLS/SSL mode" instead of implying a guided/verified step. If we want it as a real step, it needs a `data-testid` added to the Combobox in `grafana-postgresql-datasource` first.

postgresql-data-source-lj/add-data-source/content.json
lwandz13: Fixed in 4f72ab80: changed to "PostgreSQL is a built-in core data source in Grafana, so you don't need to install a plugin" — accurate across OSS, Enterprise, and Cloud.

postgresql-data-source-lj/test-connection/content.json
lwandz13: Fixed in 4f72ab80: the recommendation is now conditional — "If your PostgreSQL database isn't publicly accessible, use PDC to establish a secure connection. If your database is publicly reachable, you can skip the PDC selection in the next step." That matches the docs and lines up with the skippable PDC step.

tacole02 (approved):
Thanks for the follow-up. The four notes from the earlier review are addressed in 4f72ab80, and the previous Block Editor pass still applies.

Looks good to merge from my testing.

PR #570, opened 2026-09-09
feat(transformations): teach transformations on a data source every Grafana has

Jayclifford345:
## What this adds

`transform-testdata-query` — a four-section guide teaching data transformations on the **TestData**
data source: add TestData, build a three-series Random Walk query, apply **Reduce**, then chain
**Organize fields by name** so the reader sees that pipeline order is what matters.

## Why another transformations guide

The two that exist are both Cloud-only, and not by accident of wording — by targeting:

| guide | targeting | reachable on OSS/local? |
|---|---|---|
| `find-transformations` | `source: .*\.grafana\.net` | no — `source` is in the OSS matcher's unsupported set, so the whole entry is discarded |
| `data-transformation-lp` | `targetPlatform: cloud` | no |
| this one | bare `urlPrefixIn: ["/dashboard/new", "/d/"]` | yes, any tier |

TestData ships with every Grafana, so the query half of the lesson behaves identically everywhere
and no Prometheus is needed. The first section creates the data source if the reader has not got
one, gated on `is-admin` and skippable.

## Selectors

Reftargets are `{grafana:...}` paths wherever `@grafana/e2e-selectors` has a correct entry, so they
resolve against the reader's own Grafana instead of pinning the guide to the version it was authored
against. Three exceptions, each measured against a running instance rather than assumed:

- **The TestData card** on `/connections/datasources/new` uses the `button` action on visible text.
  `pages.AddDataSource.dataSourcePluginsV2` looks like the right anchor and *is* set in
  `DataSourceTypeCard.tsx`, but `Card.Heading` does not forward the caller's `data-testid` — all 52
  cards render `data-testid Card heading` instead, so the declared selector never reaches the DOM.
  Worth a separate fix in grafana/grafana; flagging it here rather than working around it silently.
- **`input[name='seriesCount']`** has no entry in the selector package.
- **`{grafana:components.Transforms.card:Reduce} button`** — the path resolves to a wrapper `div`
  and the click has to land on the `button` inside it, so the reftarget keeps a descendant
  combinator. Same for the datasource picker card.

## Verification

On a clean `grafana/grafana:13.1.3` container with **zero** data sources, so the first section ran
as a first-time reader meets it:

- Schema-valid against the Pathfinder MCP (schema 1.1.0), 11 top-level blocks.
- `scripts/lint-targeting.py` — not flagged.
- Walked **in order**: all 17 steps matched exactly one element at the moment their own step ran.
  End state: two transformations applied, three series reduced to three rows, and the Organize
  editor listing Reduce's output columns — which is the claim the last markdown block makes.
- `verify` on the data-source step resolves on the page the step lands on.

`minVersion: 13.1.0`, because it uses the new dashboard edit experience. Nothing here claims to work
on 12.x — it was not tested there.

## Rules checked

Against `AGENTS.md`: `exists-reftarget` on every targeting step, `on-page:` on page-specific steps,
`verify` after the create, one-sentence tooltips under 250 characters that do not name the
highlighted element, section bookends outside the sections, no markdown inside a section, no
multistep singletons, GUI names bolded and nothing else. The transformation drawer and the data
source list are both fully rendered rather than virtualised — checked by counting DOM nodes — so no
`guided`/`lazyRender` is needed.

### DIFF
--- a/transform-testdata-query/content.json
+++ b/transform-testdata-query/content.json
@@ -100,34 +100,18 @@
           ]
         },
         {
-          "type": "multistep",
+          "type": "interactive",
           "id": "choose-testdata",
-          "content": "Point the query at **TestData**.",
-          "steps": [
-            {
-              "id": "filter-datasources",
-              "action": "formfill",
-              "reftarget": "{grafana:components.DataSourcePicker.inputV2}",
-              "targetvalue": "testdata",
-              "requirements": [
-                "exists-reftarget",
-                "has-datasource:grafana-testdata-datasource"
-              ],
-              "formHint": "testdata",
-              "tooltip": "Typing narrows the list to one entry, so the next click cannot land on the wrong data source.",
-              "description": "Filter the data source list."
-            },
-            {
-              "id": "select-testdata",
-              "action": "highlight",
-              "reftarget": "{grafana:components.DataSourcePicker.dataSourceList} [data-testid='data-source-card'] button",
-              "requirements": [
-                "exists-reftarget"
-              ],
-              "tooltip": "One entry is left after filtering, and this selects it.",
-              "description": "Choose TestData."
-            }
-          ]
+          "action": "formfill",
+          "reftarget": "{grafana:components.DataSourcePicker.inputV2}",
+          "targetvalue": "TestData",
+          "requirements": [
+            "exists-reftarget",
+            "has-datasource:grafana-testdata-datasource"
+          ],
+          "formHint": "TestData",
+          "content": "Set the panel's **Data source** to **TestData**.",
+          "tooltip": "Typing filters the list and picks the match, so there is nothing left to click afterwards."
         },
         {
           "type": "multistep",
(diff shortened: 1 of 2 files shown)


transform-testdata-query/content.json
```diff
@@ -0,0 +1,286 @@
...
+    },
+    {
+      "type": "section",
+      "id": "build-query",
+      "title": "Build a query worth transforming",
+      "blocks": [
+        {
+          "type": "multistep",
+          "id": "open-panel-editor",
+          "content": "Create a panel and open the panel editor.",
```
imatwawana, line 70: This step creates a dashboard, but never explicitly calls that out.

imatwawana (commented):
So the overall content is useful, but the language is less clear than I'd want. Some of it is slightly weird sentence structures and some of it's just not in plain enough in the language (like assuming people know what a data frame is). I think the skill has generated usable content though; it would just require a solid edit of the language. Let's call this a 6.5-7/10

transform-testdata-query/content.json
Jayclifford345: Updated the opening step to explicitly create a new dashboard, add a panel, and open its editor. I also edited the guide for plain language, explained data frames and fields, and clarified the purpose and result of each section. Strict package validation passes. Pushed in 857879a4.

mariadubinets (approved):

PR #575, opened 2026-09-11
feat: add a tour of the Kubernetes Monitoring app

Jayclifford345:
A tour of the Kubernetes Monitoring app for someone whose cluster is already sending data. Three
independent sections, each opening with its own `navigate` step, so a reader can take them in any
order or stop after one.

| section | covers |
|---|---|
| `tour-overview` | the home page's three views, the data source the page is scoped to, and the standalone Alerts page |
| `tour-lists` | the column picker and usage/cost switch on the Clusters table, then the four other lists that carry the same two controls |
| `tour-cost-and-cardinality` | where spend concentrates, what it could be, and whether the metrics behind those numbers are arriving |

It stops at the pages rather than re-teaching the drilldowns — the manifest recommends **k8s-cpu**
and **k8s-mem**, which already cover namespace → workload → pod → container for CPU and memory.

## Verification

19 steps, walked **in order** in a browser against Grafana 12.3.0 with `grafana-k8s-app` 2.38.0
built from source. Every selector matched exactly one visible element at the moment its own step
ran, and each section's success criteria were asserted against the resulting page (h1, active tab,
panel titles, row counts).

Gates run locally:

- `check-batch.py` house rules — 0 errors
- Pathfinder schema validation — OK
- `validate --strict` — valid (this first failed on 10 `parent` fields left on nested
  `multistep`/`guided` steps, which the non-strict validator passes; they are stripped)
- `lint-targeting.py` — no findings

**Not verified on 13.2.1, which is what Cloud currently runs.** Worth a pass before this leaves
draft.

## Why it is shorter than it looks like it should be

Every step targets one of four plugin-owned anchors — `prometheus-picker`,
`column-picker-button`, `column-picker-menu`, `TableViewSwitch` — plus URL navigations. A larger
tour was drafted and then cut, because a live walk found the rest unreachable:

- **Everything in a table cell.** Pod-health bars, forecast buttons, manifest links and status
  links all sit in the trailing columns of a `react-data-grid` that virtualises **horizontally**,
  so they are not in the DOM until the grid is scrolled sideways. Measured 0 matches on load
  against 17 and 24 after scrolling. `lazyRender` cannot reach them — it only scrolls vertically
  (grafana/grafana-pathfinder-app#1865).
- **Dead code.** The cost breakdown switch is commented out of both cost scenes; the jobs "See all"
  button only renders when both job lists appear on one page and no caller does that.
- **A dead route pair.** `/health/risks` and `/health/efficiency` are not registered and fall
  through to the onboarding page; that content lives on the home Efficiency tab.
- **Ambiguity.** The home "Search" button has no test id and its visible text also matches the app
  nav item; `[data-testid='status-panel'] a` matches 26 Explore links.

A test-id PR against `grafana/grafana-k8s-plugin` is drafted for the four of those a test id can
actually fix. The other five need the runtime fix or a behaviour change first.

### DIFF
--- a/grafana-k8s-tour/content.json
+++ b/grafana-k8s-tour/content.json
@@ -4,34 +4,37 @@
   "blocks": [
     {
       "type": "markdown",
-      "content": "This tour has three independent sections you can take in any order \u2014 start here to see how the home page's three views divide cluster health, efficiency and alerts, and where firing alerts collect on their own page."
+      "content": "This tour walks the Kubernetes Monitoring app in five independent stops \u2014 Overview, Alerts, the Clusters list, the other object lists, and Cost \u2014 take them in any order, or stop after just one, since your cluster is already sending data."
+    },
+    {
+      "type": "markdown",
+      "content": "Here's what the home page's Overview and Efficiency views tell you about a cluster that's already reporting in."
     },
     {
       "type": "section",
       "id": "tour-overview",
-      "title": "The Kubernetes Overview and alerts",
+      "title": "The Kubernetes Overview",
       "requirements": [
-        "has-permission:grafana-k8s-app.alerts:read",
         "has-permission:grafana-k8s-app.home:read"
       ],
       "blocks": [
         {
           "type": "interactive",
-          "id": "open-overview",
+          "id": "overview-open",
           "action": "navigate",
           "reftarget": "/a/grafana-k8s-app/home/overview",
-          "content": "Open the **Overview** view \u2014 the default read on your cluster's health.",
-          "tooltip": "Cluster health, at a glance, is the first thing this app shows you.",
+          "content": "Open the Kubernetes Monitoring home page for a cluster that's already sending data.",
           "doIt": true,
-          "requirements": []
+          "requirements": [],
+          "tooltip": "The app's front door. Everything else in the tour is a narrower view of what this page summarises."
         },
         {
           "type": "interactive",
-          "id": "home-datasource",
+          "id": "overview-datasource",
           "action": "highlight",
           "reftarget": "[data-testid='prometheus-picker']",
-          "content": "Every view on this page \u2014 **Overview**, **Efficiency** and **Alerts** \u2014 scopes its panels to whichever cluster's Prometheus is picked here.",
-          "tooltip": "Every panel across all three home views reads from whichever cluster's Prometheus is chosen here.",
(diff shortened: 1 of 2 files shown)


Jayclifford345:
Rebuilt on review feedback (`630dd249`). Two things were wrong with the first draft:

- several page changes were chained inside `multistep` blocks with nothing done or explained on the
  pages they passed through
- the Prometheus data-source control was highlighted three times instead of anything specific to
  each page

Both had the same root cause: the guide was built only from anchors the plugin's own TSX declares,
and there are four of those. That left URL hops for navigation and the data-source picker as filler.

**What changed.** 22 core-Grafana anchors were added — panel headers, tabs, nav items and
template-variable labels, the same family the published `k8s-cpu` and `k8s-mem` guides already use.
Each was measured in a browser at exactly one visible match, and each cites the plugin call site
supplying the title or variable name it is keyed on.

| | before | after |
|---|---|---|
| sections | 3 | 5 |
| steps | 19 | 32 |
| page changes by click | 0 | 9 |
| data-source highlights | 3 | 1 |

Every page now carries a highlight of something specific to it: the count row and risk row on the
overview, alerts by cluster and by severity, each list's own filter variable, top spenders, CPU cost
against savings, metrics per cluster.

Section 4 ends on the **Jobs** tab rather than Cronjobs — All jobs lands on Cronjobs by default, so
clicking that tab would have been a no-op on an already-active control.

**Verification.** All 32 steps walked in order against Grafana 12.3.0 with the plugin built from
source: every selector one visible match at the moment its own step ran, every click executed and
its landing page confirmed, all 15 success criteria asserted. `check-batch` 0 errors, schema OK,
`validate --strict` valid, `lint-targeting` no findings.

Two defects the gates caught after generation, both worth knowing about:

1. `check-batch` requires a tooltip on every actioned step, including `navigate`. Added.
2. Six `verify` fields had been written as CSS selectors. `vocab.md` describes `verify` as "CSS
   selector to check for verification after action", which is wrong — the strict validator rejects
   a selector with `Unknown condition type`, and at runtime an unknown token makes the step never
   complete, leaving the reader stuck with nothing on screen to explain why. They are now
   `on-page:` tokens naming the page each click was measured to land on.

Still not verified on 13.2.1, which is what Cloud runs — that is the one thing outstanding before
this leaves draft.

mariadubinets (approved):

PR #587, opened 2026-09-14
Add guide: Build a table with filters, summaries, and gauges

Jayclifford345:
Adds “Build a table with filters, summaries, and gauges.” Readers use pagination and column filters, compare Mean and Max footer calculations, and apply gauges to the numeric field. A concrete loaded-data example explains automatic gauge bounds and why a full gauge is not necessarily 100% capacity. Row-spacing, width, header-wrapping, and cell-inspection detours are omitted.

Instructions identify the UI section, setting, and value. Section bookends stay outside the numbered steps; opening Go there actions have their own sections. Retained action IDs, selectors, values, and grouped dashboard saves are preserved. Recommendations remain limited to dashboard editing with the matching visualization tag.

Validation: strict Pathfinder package validation, ID/reference checks, retained-action checks, whitespace checks, and independent changed-guide review pass. No unresolved review threads were present when this update was prepared.

Live validation (2026-09-21, connected Chrome on learn.grafana.net): existing-source setup, panel creation, bounded samples, pagination/filters/decimals, Mean/Max footer, numeric-field gauges, all quizzes, and final inspection passed through Pathfinder. The suggested dashboard name already existed, so a separate validation copy was saved with a suffix. Reload and visible output confirmed the configuration persisted. Fresh TestData creation was separately replayed through the documented UI and Save & test reported working. The final opening-section restoration and timestamp-label correction preserve the tested actions.


grafana-table-signal-workbench/content.json
```diff
@@ -0,0 +1,827 @@
+{
+  "schemaVersion": "1.1.0",
+  "id": "grafana-table-signal-workbench",
+  "title": "Signal workbench: advanced table techniques",
```
imatwawana, line 4: This title seems a bit obscure

grafana-table-signal-workbench/content.json
```diff
@@ -0,0 +1,827 @@
...
+  "blocks": [
+    {
+      "type": "markdown",
+      "content": "Start in **Dashboards** and build a standalone table with synthetic signals, readable rows, summary calculations, and numeric gauges."
+    },
+    {
+      "type": "section",
```
imatwawana, line 11: A bit odd that it's got two one-step sections

grafana-table-signal-workbench/content.json
```diff
@@ -0,0 +1,827 @@
...
+      "requirements": [
+        "is-admin"
+      ],
+      "objectives": [
+        "has-datasource:grafana-testdata-datasource"
+      ]
+    },
+    {
+      "type": "markdown",
+      "content": "You have a synthetic source to practice with without external services or credentials."
```
imatwawana, line 105: Fancy!

grafana-table-signal-workbench/content.json
```diff
@@ -0,0 +1,827 @@
...
+      "type": "markdown",
+      "content": "You have a named panel connected to the source chosen for this sitting."
+    },
+    {
+      "type": "markdown",
+      "content": "Generate one random walk from 50 within a range of 0 to 100 so you can focus on reading the table."
+    },
+    {
+      "type": "section",
+      "id": "create-bounded-practice-signals",
```
imatwawana, line 251: The skill doesn't always use the language in the UI consistently in this section, so if you don't catch where it made a change after clicking Do it you end up having to go back and do a Show me. I think maybe guiding the skill to use UI strings as part of its directions would be good.

imatwawana (commented):
Having done a couple of these signal ones, I like how they explain how these manipulations of the visualization options. I think the overall idea is a good one. This one just doesn't consistently use UI language, which I think would be a worthy improvement to the skill.

grafana-table-signal-workbench/content.json
Jayclifford345: Renamed the guide to “Build a table with filters, summaries, and gauges” and updated its manifest description. Pushed in 0f64d836.

grafana-table-signal-workbench/content.json
Jayclifford345: Removed the opening single-action section and split optional TestData setup into its individual actions. Omitted the redundant Save & test action because selecting TestData creates the source immediately. Pushed in 0f64d836.

grafana-table-signal-workbench/content.json
Jayclifford345: Kept the data-source selection approach in place; no change needed here.

grafana-table-signal-workbench/content.json
Jayclifford345: Updated the guide instructions to name the UI sections, controls, and values: Table, Standard options, Table footer, Cell options, and override settings. Also matched the current Add/Panel, Edit visualization, All visualizations, Back, and Save labels. Pushed in 0f64d836.

mariadubinets (approved):

PR #588, opened 2026-09-14
Add advanced Stat Signal snapshot walkthrough

Jayclifford345:
Adds a standalone Stat guide comparing individual samples with a calculated mean, then using a name, sparkline, threshold background, and percent change. A worked example explains first-to-last non-null percent change separately from the displayed mean; inverted coloring is explained before its knowledge check. Cosmetic layout exercises and the synthetic-data quiz are omitted.

Instructions identify the UI section, setting, and value. Section bookends stay outside the numbered steps; opening Go there actions have their own sections. Retained action IDs, selectors, values, and grouped dashboard saves are preserved. Recommendations remain limited to dashboard editing with the matching visualization tag.

Validation: strict Pathfinder package validation, ID/reference checks, retained-action checks, whitespace checks, and independent changed-guide review pass. No unresolved review threads were present when this update was prepared.

Live validation (2026-09-21, connected Chrome on learn.grafana.net): restored opening section, existing-source setup, panel/query creation, sample-versus-mean lesson, sparkline/background settings, thresholds/percent change, both quizzes, grouped save and final inspection passed through Pathfinder. Reload confirmed the named mean and percent-change display persisted. Fresh TestData creation was separately replayed through the documented UI and Save & test reported working.


grafana-stat-signal-snapshot/content.json
```diff
@@ -0,0 +1,776 @@
...
+          "action": "formfill",
+          "reftarget": "{grafana:components.DataSource.TestData.QueryTab.seriesCount}:is(input)",
+          "requirements": [
+            "exists-reftarget",
+            "on-page:/dashboard/new"
+          ],
+          "tooltip": "Keep the exercise focused on one signal.",
+          "targetvalue": "1",
+          "type": "interactive",
+          "content": "Set **Series count** to 1.",
```
imatwawana, line 306: I don't think you ever need a step to add one series. There's always at least one series.

imatwawana (commented):
Just a nit I noticed on all of them

grafana-stat-signal-snapshot/content.json
Jayclifford345: Removed the redundant Series count = 1 step; the Stat guide now uses the default single series. I also checked the related Table and Bar chart guides: Table receives the same cleanup, while Bar chart needs its explicit count of 3. Strict package validation passes. Pushed in 97c97919.

mariadubinets (approved):

Issue #513, opened 2026-08-13
Suppress screenshots and videos in Pathfinder learning paths


tacole02:
## Goal

Suppress **all screenshots and videos** from the Pathfinder rendering of learning paths. Website rendering can keep them.

## Scope

- Full audit of learning path packages in this repo
- Remove or gate media so Pathfinder does not show screenshots/videos
- Leave screenshots/videos intact for website rendering where they already exist

## Acceptance criteria

- [ ] Audit all learning path packages for screenshots and videos
- [ ] Pathfinder (app) versions no longer surface screenshots or videos
- [ ] Website rendering still shows screenshots/videos as today
- [ ] Document the pattern used (e.g. website-only fields, conditional blocks) so future paths stay consistent

## Notes

- Related: Visualize metrics, logs, traces was already flagged as having screenshots that should be removed

PR #591, opened 2026-09-14
Enforce website-only media gating in create/build/review/preflight LP…

### LINKED ISSUE #562: Enforce website-only media gating in create, review, and preflight LP skills
## Goal

Bake the Pathfinder website-only media pattern from #518 / #513 into the learning path authoring and review skills so new and reviewed packages don't ship ungated screenshots/videos.

## Context

Pathfinder no longer shows screenshots or embedded videos in learning paths; the website still does. Media must be wrapped in:

```json
{
  "type": "conditional",
  "conditions": ["renderer:website"],
  "whenTrue": [ /* image or video block */ ],
  "whenFalse": []
}
```

Markdown that mixes prose with `` uses a dual-branch conditional (`whenTrue` = with images; `whenFalse` = prose-only).

Docs: `docs/learning-path-authoring.md` (Screenshots and videos), `docs/requirements-reference.md`, `docs/json-guide-reference.md`.

## Scope

Update these so they teach, generate, and check the pattern:

1. **Create path** — [`.cursor/commands/create-learning-path.md`](.cursor/commands/create-learning-path.md) and related create-learning-path references / workflows
2. **Convert / build** — [`.cursor/commands/build-interactive-lj.md`](.cursor/commands/build-interactive-lj.md) (if media is copied from website markdown during convert)
3. **Review path** — [`.cursor/skills/review-learning-path/SKILL.md`](.cursor/skills/review-learning-path/SKILL.md) and shared [learning-hub-standards.md](.cursor/skills/review-learning-path/learning-hub-standards.md) / reference-checks (flag ungated media as a finding)
4. **Preflight path** — [`.cursor/skills/preflight-learning-path/SKILL.md`](.cursor/skills/preflight-learning-path/SKILL.md) and its reference-checks (same ungated-media check from the author side)

Also confirm [migrate-guide](.cursor/skills/migrate-guide/SKILL.md) still points authors at wrap-not-strip for absolute media URLs.

## Acceptance criteria

- [ ] Create / convert workflows produce `renderer:website` wrappers for any new `image` / `video` / markdown embeds
- [ ] Review and preflight skills flag ungated `image`, `video`, or markdown `` as actionable findings
- [ ] Dual-branch rule for mixed prose + images is documented in the skill checklists (not only in learning-path-authoring.md)
- [ ] Plain YouTube text links and `website.yaml` `cta.image` remain out of scope for the gate

## Out of scope

- Website Hugo renderer changes for `conditional` (tracked separately in website)
- Retroactively re-auditing all packages already gated by #518

tacole02:
Bakes the renderer:website conditional pattern from #513/#518 into the two authoring entry points that create new media-bearing content (create-learning-path, build-interactive-lj) and fixes the finding routing on the review/preflight side, where "ungated media present" was conflated with the soft "no screenshot yet" note and silently suppressed instead of surfaced to the author.

Closes #562


moxious (changes requested):
**Requesting changes** — found one blocking contradiction and a few consistency gaps that should be resolved before merge.

### Blocking

**1. `.cursor/commands/build-interactive-lj.md:21`** — Step 3 instructs gating a carried-over screenshot with a `conditional` block, but this directly contradicts the business-value milestone's markdown-only rule at line 38 ("always uses markdown blocks exclusively"). Since `conditional` is called out elsewhere in this file as a distinct block type, and it's not addressed either way in the forbidden-block list, an agent following this doc has no way to resolve whether wrapping business-value content in `conditional` is allowed. Please clarify which rule wins, or add an explicit carve-out.

### Should fix

**2. Missing YouTube/`cta.image` exception in `build-interactive-lj.md`** — `create-learning-path.md:24` explicitly exempts plain YouTube text links and `website.yaml`'s `cta.image` from gating, but `build-interactive-lj.md` (Step 3, Critical Rule 4, and the anti-pattern bullet at line 59) never mentions this exception. This file is specifically used to migrate *existing* learning paths — exactly where legacy YouTube links and `cta.image` are most likely to already exist — so as written it risks incorrectly wrapping valid ungated content.

**3. `review-learning-path/reference-checks.md:29`** — same asymmetry in the other direction: the reviewer-facing checklist doesn't restate the exception that `preflight-learning-path/reference-checks.md:73` states explicitly. Low risk in practice (given where `cta.image` lives), but worth aligning the two checklists' documented criteria.

**4. `review-learning-path/learning-hub-standards.md:311`** — This PR adds a paragraph + table row (~line 71-84) distinguishing "no screenshot" (OK) from "screenshot present but ungated" (must-fix) — the exact bug from #562. But the separate "Common pitfalls" quick-scan table wasn't updated with a row for the ungated case, so a reviewer using only that table can still miss the finding this PR exists to catch.

### Consider

**5. `review-learning-path/reference-checks.md:28`** — The fix splits the one conflated cell that caused #562, but doesn't add a general "one finding per cell, never bundle findings of differing severity" rule. Other cells (e.g. line 27) still bundle distinct checks, so the same silent-suppression pattern can recur elsewhere.

**6. Duplication** — The full media-gating recipe (whenTrue/whenFalse/dual-branch mechanics) is now duplicated near-verbatim across `build-interactive-lj.md`, `create-learning-path.md`, and `migrate-guide/SKILL.md`, even though all three already link to `docs/learning-path-authoring.md#screenshots-and-videos-website-only-in-pathfinder`. Finding #2 is direct evidence this is already causing drift — consider centralizing behind that doc link instead of restating the mechanics inline.

tacole02:
@moxious thanks for the thorough review — addressed all six points in 63492d8:

**Blocking**
1. Resolved the `business-value`/`conditional` conflict: Critical Rule 5 in `build-interactive-lj.md` now explicitly allows a `conditional` block when both `whenTrue` and `whenFalse` contain markdown only, since that doesn't add an interactive/section/guided/multistep block. Step 3 links back to the rule for the carve-out.

**Should fix**
2. Added the YouTube-link / `cta.image` exception to `build-interactive-lj.md` (Step 3 + Critical Rule 4), matching `create-learning-path.md`.
3. Restated that same exception in `review-learning-path/reference-checks.md`'s milestone checklist row.
4. Added an "Ungated media visible in Pathfinder" row to the `learning-hub-standards.md` "Common pitfalls" quick-scan table, not just the Landing screenshot section.

**Consider**
5. Added a "One finding per cell" rule to the finding-routing table in `reference-checks.md`, calling out that this exact bundling bug is why #562 needed a fix, so reviewers don't reintroduce it elsewhere.
6. Trimmed the duplicated wrap/dual-branch mechanics in `build-interactive-lj.md` and `create-learning-path.md` down to a link to `docs/learning-path-authoring.md` — kept the hazard-callout exceptions inline (since dropping those is exactly what caused #2) but removed the restated mechanics that were drifting. Left `migrate-guide/SKILL.md` as-is since it predates this PR and carries its own migration-notes audit-trail requirement beyond the base recipe.

Would appreciate another look when you have a chance.

.cursor/commands/build-interactive-lj.md
```diff
@@ -34,8 +34,8 @@ For background on how this command relates to `/create-learning-path`, refer to
...
+4. **Capture all content in `content.json`.** Since the website no longer renders this learning path, every milestone's conceptual prose must be carried into its `content.json` as markdown blocks. Nothing should remain only in the website markdown. Gate any carried-over screenshot, video, or markdown image per [Screenshots and videos](../../docs/learning-path-authoring.md#screenshots-and-videos-website-only-in-pathfinder) — don't leave it ungated just because it came from the website source. Don't gate plain YouTube text links or `website.yaml` `cta.image`.
+5. **`business-value` is markdown-only.** The `business-value` milestone always uses markdown blocks exclusively. Never add interactive, section, guided, or multistep blocks to it. A `conditional` block is allowed **only** when both `whenTrue` and `whenFalse` contain markdown blocks and nothing else (for example, gating a screenshot/video embedded in its markdown per Rule 4) — that doesn't violate the markdown-only rule because neither branch adds an interactive, section, guided, or multistep block.
```
moxious: The new carve-out only covers the dual-branch case (`whenTrue`/`whenFalse` both markdown). It doesn't cover the dedicated image/video wrap (`whenTrue: [{"type": "image"}]`, `whenFalse: []`), which is actually the more common pattern in existing business-value milestones today (`visualization-metrics-lj`, `visualization-traces-lj`, `google-sheets-data-source-lj` all use it; only `visualization-logs-lp` uses the dual-branch form this rule describes). As written, an agent following this rule would flag three of the four existing precedents as non-compliant. Suggest widening it to something like "every block in `whenTrue`/`whenFalse` is `markdown`, `image`, or `video`, never `interactive`, `section`, `guided`, or `multistep`."

moxious (changes requested):
Thanks for the update. The business-value/conditional contradiction, the YouTube and `cta.image` exceptions, the checklist asymmetry, and the missing pitfalls-table row are all resolved.

I left one note on the diff: the new Rule 5 wording for the conditional carve-out is narrower than the actual gating pattern in use, and would contradict several existing business-value milestones that gate a screenshot or video directly rather than embedding it in markdown.

.cursor/commands/build-interactive-lj.md
tacole02: Good catch — you're right, I checked `visualization-metrics-lj`, `visualization-traces-lj`, and `google-sheets-data-source-lj` and confirmed all three use the dedicated `image`/`video` wrap (`whenTrue: [{"type": "image"}]`, `whenFalse: []`), not the dual-branch markdown form. Widened Rule 5 in fa5a04a8: it now permits `markdown`, `image`, or `video` anywhere in `whenTrue`/`whenFalse`, still forbidding `interactive`/`section`/`guided`/`multistep` there. That covers both existing patterns.

moxious (approved):

PR #592, opened 2026-09-15
Add guide: Build a dashboard with dependent filters

Jayclifford345:
Adds one standalone guide that installs QuickPizza when needed and builds a separate dashboard with dependent environment/service query variables. Readers configure refresh and multi-value/All interpolation, deliberately introduce a hardcoded filter, inspect and repair the query, compare service selections, then save and reopen the dashboard.

Installation records progress before the demo plugin reloads Grafana, and the next section requires the demo data source. Variable creation opens the Add pane idempotently and exposes the type/name actions individually. Broad panel highlights are replaced by concise result checks; query drawers and dashboard saving remain grouped. The final dashboard search result uses guided lazy rendering.

The guide uses `grafanacloud-play-prom` and QuickPizza metrics. Only the production environment was observed; the guide does not invent additional environments and explains that dashboard variables do not apply to alert rules.

Validation: strict package validation, ID/reference checks, and changed-block review pass. Targeted Cloud testing on learn.grafana.net verified the existing-install objective, the complete eight-step environment and seven-step dependent-service sections through Pathfinder Run section, including query previews, Apply, refresh, and selection settings. The resulting selections were production and quickpizza-catalog. The uninstalled QuickPizza branch was checked against the installer source and the repository's reload-safe pattern, but was not freshly installed on this already-configured shared instance. This is not a complete rerun of the entire final guide.

To try it, import `grafana-quickpizza-working-filters/content.json` into Pathfinder and select Preview on a Grafana instance with Demo Data Dashboards and the new dashboard editor.


grafana-quickpizza-working-filters/content.json
```diff
@@ -0,0 +1,1205 @@
...
+            "on-page:/a/grafana-demodashboards-app",
+            "min-version:13.3.0"
+          ],
+          "type": "interactive",
+          "content": "Install QuickPizza, or open its dashboards when already installed",
+          "doIt": true,
+          "showMe": true,
+          "verify": "on-page:/dashboards/f/grafana-demodashboards-quickpizza",
+          "tooltip": "Install QuickPizza, or open its dashboards when already installed."
+        }
```
leslie-heinzen, line 45: This step gets stuck if quickpizza isn't installed yet, might be a requirements check thing

https://github.com/user-attachments/assets/dacb7cb2-4984-498e-95d4-4fe02692a20b

grafana-quickpizza-working-filters/content.json
```diff
@@ -0,0 +1,1205 @@
...
+              "action": "highlight",
+              "reftarget": "{grafana:components.PanelEditor.ElementEditPane.variableType:query} button",
+              "requirements": [
+                "exists-reftarget",
+                "on-page:/dashboard/new",
+                "min-version:13.3.0"
+              ],
+              "description": "Choose Query",
+              "tooltip": "Choose Query."
+            },
```
leslie-heinzen, line 101: This substep failed when I test it. Weirdly the next step that involved adding a query variable worked fine ("Add a Query variable named service."); maybe double check the steps involved there against these.

https://github.com/user-attachments/assets/a272596c-0971-4c3a-930c-6d5e679aa102

leslie-heinzen (commented):
Couple of steps early in the guide that hung but beyond that it worked smoothly enough.

There are some "highlight" steps in here that I think aren't really adding anything beyond being readonly steps. Namely steps that highlight the entire panel and instruct the user to check it; imo those aren't super useful. 

The guide is kinda hard to follow for me because the guide text is so sparse; just bare descriptions of the actions taken. But I got kinda lost with what we were trying to do in the guide halfway through.

Heavy on the multisteps; kind of masks what the guide is going to do for a given step. Maybe good from a "do it for me" angle, but maybe not super educational.

grafana-quickpizza-working-filters/content.json
Jayclifford345: Fixed the installation flow in 4b9db0b5: the demo installer reloads Grafana, so the step now records completion before navigation and the section auto-completes only when the demo data source exists. The next section also requires that data source. Verified the already-installed branch on learn.grafana.net. The clean-install branch follows the repository reload-safe pattern and installer source; I did not remove the existing shared demo to retest installation.

grafana-quickpizza-working-filters/content.json
Jayclifford345: Added an explicit idempotent Add-pane action before the first variable and split variable type/name setup into separate actions. The final eight-step environment section passed through Pathfinder Run section on learn.grafana.net, including query preview, Apply, refresh settings, and multi-value/All options. Also clarified the learning goal and replaced broad panel highlights with concise result checks. Pushed in 4b9db0b5.

mariadubinets (approved):

PR #602, opened 2026-09-18
Fix billing and usage alert guide flows

leslie-heinzen:
## Context

This PR repairs two related but distinct CMAB UI regressions across three billing-alert guides.

### Onboarding milestone

`grafana-cloud-onboarding-lj/setup-billing-alerts` started failing between its last observed pass on September 4 and its first failure on September 17:

```text
Element not found: All Products
```

[grafana-cmab-app#2023](https://github.com/grafana/grafana-cmab-app/pull/2023) ([commit `0be38b02`](https://github.com/grafana/grafana-cmab-app/commit/0be38b02)) replaced the persistent alert starter cards with a recommendation card for the affected self-serve state. **All Products** remained available only inside the transient **Create alert** menu; the removed card had also provided a fallback target for the guide's text lookup.

The guide opened **Create alert** and selected **All Products** as separate steps. Clicking the second **Do it** control occurs outside the custom dropdown, which dismisses the menu before the guide resolves **All Products**.

### Existing usage-alert guides

`create-usage-alert` and `billing-usage-lj/create-billing-alert` were authored against the previous direct **Create usage alert** button in June. That label is present in the CMAB source before July 2.

[grafana-cmab-app#1702](https://github.com/grafana/grafana-cmab-app/pull/1702) ([commit `665de6a8`](https://github.com/grafana/grafana-cmab-app/commit/665de6a831b95d3d5b9392dbf13d4c7059cf1712)) removed the old `Create usage alert` label and introduced the **Create alert** menu. The two guides still targeted the old control, so their opening step skipped and the following mandatory drawer step failed.

The current builder also requires a threshold, at least one alert level, and a contact point. Contact-point availability is stack-specific, so guides should not represent selection or submission as reliably executable actions.

## Changes

- Use a single `multistep` sequence to click **Create alert** and select **All Products** in all three guides.
- Select **50% of threshold** in the onboarding milestone, which previously omitted an alert-level action.
- Make contact-point selection and final submission manual markdown instructions where applicable.
- Replace the onboarding milestone's final external documentation action with a markdown link.

## Validation

- validated all three packages with `pathfinder-cli validate --package`
- tested the local guides against a disposable Grafana Cloud development stack with the current CMAB UI
- onboarding milestone: **7 passed, 0 failed, 0 skipped**
- related guides: **2/2 guides passed, 13/13 steps passed**
- confirmed that each multistep opens the drawer with **Global Spend** selected
- confirmed that the onboarding milestone selects the 50% alert level

The tests did not submit alerts because the stack had no configured contact points and submission is intentionally manual. No later onboarding milestone requires a created alert.


violetadev:
Thanks for tracking this down! The guide fixes look right to me (routing both steps through a single `multistep` against the transient **Create alert** menu works in either UI state, so it's robust regardless of which one your stack serves). 

Two corrections to the **Context** section though.

**1. `#2023` has not been released.**

`grafana-cmab-app` publishes to `ops`/`prod` only when `package.json`'s version changes on `main` — the `check-version` gate in `publish.yml`. 

The version has been `1.28.8` since 2026-08-20, so the publish run for `0be38b02` was a no-op (`version-changed=false`, `cd` job skipped) and `git tag --contains 0be38b02` is empty. The catalog is still on v1.28.8.

**What every `main` push *does* trigger is `publish-dev.yml`, which deploys to `dev` only. So `0be38b02` is live on dev and nowhere else.**

That matters because the change is itself dev-gated. `selfServeUsageAlertsConfig` is wrapped in `whenLocalOrDev`, which falls back to `usageAlertsDefaultConfig` (`STARTER_CARDS`) on `grafana.com` / `grafana-cloud.com`. So:

- on a **dev** stack (`*.grafana-dev.net`, demokit), a self-serve org now gets `RecommendedAlertCards` — the persistent **All Products** card is gone. This is your failure, and `0be38b02` is correctly identified as the cause.
- on **prod/ops**, self-serve orgs still get the starter cards, including **All Products** — both today and after v1.28.8's successor ships.

**Could you confirm which environment the failing milestone runs against?** 

If it's prod, `0be38b02` isn't the cause and something else regressed between Sept 4 and Sept 17. If it's dev, the attribution is right but worth saying so explicitly, since the description currently reads as a prod regression.

Also note `0be38b02` merged at 15:22 BST on Sept 17, so depending on when the milestone ran that day it may not have been deployed yet.

leslie-heinzen:
@violetadev Thanks - we run e2e tests in dev. This draft PR is preparatory 😸

